PR #6, opened 2019-01-15
Get CI running

lalexgap:
Added a basic CI config file to root that should (hopefully) run `test:ci` on all projects. We'll probably want to update this with some smarts to only test what has been changed later on.


packages/tictactoe/src/redux/game/reducer.ts
```diff
@@ -10,7 +10,7 @@ import { LoginSuccess, LOGIN_SUCCESS } from '../login/actions';
 
 import hexToBN from '../../utils/hexToBN';
 import bnToHex from '../../utils/bnToHex';
-import { RESTING } from 'src/core/positions';
+import { RESTING } from '../../core/positions';
```
kerzhner: @lalexgap Curious, do you know why imports like `src/core/positions` do not work?

PR #7, opened 2019-01-15
Split out wallet

lalexgap:
Split out the wallet so it can be hosted in its own iFrame.


packages/wallet-comm/src/interface/to-wallet.ts
```diff
@@ -60,14 +67,13 @@ export const closeChannelRequest = () => ({
 export type CloseChannelRequest = ReturnType<typeof closeChannelRequest>;
 
 
-// VALIDATION
+// VAßLIDATION
```
kerzhner:
```suggestion
// VALIDATION
```

packages/wallet-comm/src/wallet-functions.ts
```diff
@@ -0,0 +1,163 @@
...
+  iFrame.style.display = 'none';
+  iFrame.style.position = 'absolute';
+  iFrame.style.left = '0px';
+  iFrame.style.right = '0px';
+  iFrame.style.bottom = '0px';
+  iFrame.style.top = '0px';
+  iFrame.width = '0';
+  iFrame.height = '0';
+  iFrame.style.zIndex = '9999';
+  iFrame
```
kerzhner: Is ts linting enabled for this code?

packages/wallet-comm/src/wallet-functions.ts
```diff
@@ -0,0 +1,163 @@
...
+
+}
+/**
+ *  Validates signed data.
+ * Promise resolves when the data is verified or an error occurs.
+ */
+export async function validateSignature(iFrameId: string, data, signature: string): Promise<boolean> {
+  const iFrame = <HTMLIFrameElement>document.getElementById(iFrameId);
+  const message = validationRequest(data, signature);
+  iFrame.contentWindow.postMessage(message, "*");
```
kerzhner: Can there exist a race condition where the response event is fired before the listener is added?

packages/wallet-comm/src/wallet-functions.ts
lalexgap: Not yet, that's still a TODO (This one particular case has been cleaned up)

packages/wallet-comm/src/wallet-functions.ts
lalexgap: That's a good point, we should probably create the promise before posting the message

PR #9, opened 2019-01-16
Readme

kerzhner:
This is the set of commands that worked for me, but please correct these if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,9 @@
+# apps
+Proof of concept applications that use the Force Move Game framework.
+
+### Setting up development environment and running the app
+* Build `wallet-comm` package via `yarn build` in the package directory.
+* Install lerna via `npm i -g lerna`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -0,0 +1,9 @@
+# apps
+Proof of concept applications that use the Force Move Game framework.
+
+### Setting up development environment and running the app
+* Build `wallet-comm` package via `yarn build` in the package directory.
+* Install lerna via `npm i -g lerna`
```
lalexgap: You should be able to just run `yarn` from the root directory (since the root `package.json` has an entry for `lerna`)

README.md
```diff
@@ -0,0 +1,9 @@
+# apps
+Proof of concept applications that use the Force Move Game framework.
+
+### Setting up development environment and running the app
+* Build `wallet-comm` package via `yarn build` in the package directory.
```
lalexgap: I think this also occurs during the `npx lerna bootstrap`

PR #13, opened 2019-01-18
Change truffle tests to just use ether.js

lalexgap:
Change truffle tests to use ethers.js instead. Addresses https://zube.io/magmo/client-wallet/c/380


packages/rps-poc/README.md
```diff
@@ -71,7 +71,7 @@ This will fire up the Storybook panel inside a browser.
 
 #### To run smart contract tests:
 
-`yarn run test:truffle`
+`yarn run test:integration`
```
kerzhner: @lalexgap Do you think we should change the naming scheme? If I understand correctly, these aren't really integration tests for the app. They are integration tests for the smart contracts. So maybe `yarn run test:contracts`?

lalexgap: @kerzhner that makes sense to me.

Issue #20, opened 2019-01-18
Handle when a user declines funding


lalexgap:
Currently we offer the user the option to approve/decline funding but we don't handle it gracefully.

When a user declines funding we should probably show a message and then dump them back in the lobby.

We should also let the other user know so they're not waiting indefinitely for funding

PR #31, opened 2019-01-21
Handle when Funding is declined by a player

lalexgap:
Addresses #20


packages/wallet-client/src/interface/from-wallet.ts
```diff
@@ -11,10 +11,11 @@ export const fundingSuccess = (channelId, position: string) => ({
   channelId,
   position,
 });
-export const fundingFailure = (channelId, reason) => ({
+export const fundingFailure = (channelId: any, reason: 'FundingDeclined' | 'Other', error?: string) => ({
```
tomclose: Does the `| 'Other'` here have any effect?

lalexgap: It should limit `reason` to be either the string `Other` or `FundingDeclined`. It makes it easier to check for specific types of failures instead of trying to parse  the error message.

PR #41, opened 2019-01-22
Fix insufficient funds scenario

lalexgap:
Addresses #26 

Properly initiate close and withdraw logic when one user runs out of funds.

Also removed some unused code.


kerzhner:
Looks good! 🌋

Issue #40, opened 2019-01-22
rps-poc -> rps


kerzhner:
We should probably get rid of the poc suffix in all references to rps

PR #50, opened 2019-01-23
rename rps-poc to rps

lalexgap:
Addresses #40


packages/rps/contracts/artifacts/Migrations.json
```diff
@@ -74,9 +74,9 @@
...
   "source": "pragma solidity ^0.4.23;\n\ncontract Migrations {\n  address public owner;\n  uint public last_completed_migration;\n\n  constructor() public {\n    owner = msg.sender;\n  }\n\n  modifier restricted() {\n    if (msg.sender == owner) _;\n  }\n\n  function setCompleted(uint completed) public restricted {\n    last_completed_migration = completed;\n  }\n\n  function upgrade(address new_address) public restricted {\n    Migrations upgraded = Migrations(new_address);\n    upgraded.setCompleted(last_completed_migration);\n  }\n}\n",
-  "sourcePath": "/Users/alexgap/code/rps-poc/contracts/Migrations.sol",
+  "sourcePath": "/Users/alexgap/code/rps/contracts/Migrations.sol",
```
kerzhner: @lalexgap Do you know if the absolute paths (with your mac username hardcoded) are of importance?

lalexgap: @kerzhner @geoknee  Hmm first time I noticed these since they're auto generated by truffle. 
@geoknee The reason these artifacts are checked in is so we can point the app at a test network without having to deploy there (the artifacts specify where the contract is deployed on chain).
I think that's worthwhile but we could probably sanitize them a bit if we wanted. I think we're only reason using the `ABI`, `bytecode` and `network` sections of the JSON so the rest could be scrubbed.

Issue #34, opened 2019-01-22
Add landing / splash screen for Wallet


lalexgap:
We should add some kind of splash screen to the wallet app so if someone goes to wallet.magmo.com they see something.

PR #52, opened 2019-01-23
Added simple landing page for the wallet.

lalexgap:
Addresses #34.

Added a pretty basic wallet landing page so if people go directly to wallet.magmo.com they see something.

image


geoknee:
Looks great!

In the future we will want to harmonise the styling of this page with that of the rest of the wallet.

Issue #51, opened 2019-01-23
Get storybook working in the wallet


lalexgap:
Currently fails with this error:
```
ERROR in ./node_modules/xmlhttprequest/lib/XMLHttpRequest.js
Module not found: Error: Can't resolve 'child_process' in '/Users/alexgap/code/apps/packages/wallet/node_modules/xmlhttprequest/lib'
 @ ./node_modules/xmlhttprequest/lib/XMLHttpRequest.js 15:12-36
 @ ./node_modules/ethers/utils/web.js
 @ ./node_modules/ethers/utils/index.js
 @ ./src/utils/transaction-generator.ts
 @ ./src/integration-tests/test-utils.ts
 @ ./src/redux/reducers/__tests__/test-scenarios.ts
 @ ./src/__stories__/index.tsx
 @ ./.storybook/config.js
 @ multi ./node_modules/@storybook/core/dist/server/config/polyfills.js ./node_modules/@storybook/core/dist/server/config/globals.js ./.storybook/config.js (webpack)-hot-middleware/client.js?reload=true

ERROR in ./node_modules/xmlhttprequest/lib/XMLHttpRequest.js
Module not found: Error: Can't resolve 'fs' in '/Users/alexgap/code/apps/packages/wallet/node_modules/xmlhttprequest/lib'
 @ ./node_modules/xmlhttprequest/lib/XMLHttpRequest.js 16:9-22
 @ ./node_modules/ethers/utils/web.js
 @ ./node_modules/ethers/utils/index.js
 @ ./src/utils/transaction-generator.ts
 @ ./src/integration-tests/test-utils.ts
 @ ./src/redux/reducers/__tests__/test-scenarios.ts
 @ ./src/__stories__/index.tsx
 @ ./.storybook/config.js
 @ multi ./node_modules/@storybook/core/dist/server/config/polyfills.js ./node_modules/@storybook/core/dist/server/config/globals.js ./.storybook/config.js (webpack)-hot-middleware/client.js?reload=true
```

PR #53, opened 2019-01-24
Get storybook working

lalexgap:
Addresses #51


geoknee:
This looks good to me, @lalexgap !

I would like to understand *why* these changes fix the error in https://github.com/magmo/apps/issues/51 ? The only functional difference I can see is that `padBytes32` is now *not* used to construct the `buyIn` in `test-scenarios.ts`. And of course not needing the `wallet: ` in the store function.

Also: you do not need a review to merge this in (you may want to reinstate protection onto `master` if you have not done so already).

kerzhner:
@lalexgap @geoknee I reinstated branch protection for `master` for required approval. Do we want to block merges if CI builds are failing?

geoknee:
Re: CI fails blocking merges.

My vote would be for yes, we should block unless CI passes. 

Downsides: may have to wait a little longer for tests to run.
Upsides: much harder to introduce a breaking change into master.

kerzhner:
@geoknee sounds good. Blocking on CI tests has been enabled.

PR #54, opened 2019-01-24
[tictactoe] Increase turnNum correctly when playing again in 'loser's' app

geoknee:
Loser := whomever did not play last when a conclusive (win/lose/draw) state is reached.

### DIFF
--- a/packages/tictactoe/src/redux/game/__tests__/player-b.test.ts
+++ b/packages/tictactoe/src/redux/game/__tests__/player-b.test.ts
@@ -1,4 +1,4 @@
-import { gameReducer } from "../reducer";
+import { gameReducer, youWentLast } from "../reducer";
 import {
   Player,
   scenarios,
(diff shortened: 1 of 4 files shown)


packages/tictactoe/src/redux/game/__tests__/player-b.test.ts
```diff
@@ -256,4 +256,23 @@ describe("player B's app", () => {
...
+    });
+
+    describe("if the player decides to continue", () => {
+      const action = actions.playAgain();
+      const updatedState = gameReducer({ messageState, gameState }, action);
+
+      itIncreasesTurnNumBy(1, { gameState, messageState }, updatedState);
+      itSends(scenarios.swapRoles.resting2, updatedState);
+      itTransitionsTo(state.StateName.WaitToPlayAgain, updatedState);
+      // TODO check that whomever did not play last sends resting
```
kerzhner: @geoknee Should this TODO get resolved in this PR?

geoknee: Good point. I resolved it, and commit message has some ideas of a better approach in the future. (This PR was about fixing the reducer, rather than the tests, so I won't implement those ideas here).

Issue #21, opened 2019-01-18
Document the `wallet-client` interface


lalexgap:
We should add solid documentation for the `wallet-client` interfaces.

We could use jsdoc to generate html documentation from the interface comments.

PR #60, opened 2019-01-24
Add documentation and generate it with typedocs

lalexgap:
Addresses #21 .

I've added documentation to the wallet-client interface and generated it into a `docs` folder in wallet-client.

I've also renamed `from-wallet` to `wallet-events` and `to-wallet` to `wallet-messages` to be more in line with the pattern of the interface we're exposing.

We could probably do a second pass of `wallet-events` and rejig it so we expose types/interfaces of actions instead of relying on `ReturnType` so we can better document them.


kerzhner:
@lalexgap Do you think the message and event terminology is a bit confusing? To me, it seems like either `wallet-event` or `wallet-message` is something that can be sent to the wallet or received from the wallet. I wonder if there is more explicit terminology we could use?

lalexgap:
@kerzhner I'm open to better terminology! 

However the typical consumer `wallet-client` should never have to know about `wallet-messages` as they use the `wallet-functions` to interact with the wallet (thats why I hid the `wallet-messages` from the documentation). So I think the combo of `wallet-functions`, `wallet-event-listener` and `wallet-events` makes sense in that context.

Maybe we could rename `wallet-messages` to something more explicit like `messages-to-wallet`?

kerzhner:
That reasoning makes sense to me. Also, I like `messages-to-wallet`, it is pretty clear.

packages/wallet-client/src/wallet-event-listener.ts
```diff
@@ -1,8 +1,10 @@
 
-import { ResponseActionTypes as WalletEventTypes, ResponseAction as WalletEvent } from './interface/from-wallet';
 import { EventEmitter2 } from 'eventemitter2';
+import { WalletEventType, WalletEvent } from './wallet-events';
 
-// TODO: Type this so that we can specify the events that will be returned
+/**
+ * A wallet event listener that be used to listen to the various [[WalletEvent]] that occur in the wallet.
```
kerzhner:
```suggestion
 * A wallet event listener that can be used to listen to the various [[WalletEvent]] that occur in the wallet.
```

PR #67, opened 2019-01-25
WIP: Validate Transactions by using read only call

lalexgap:
This is related to https://zube.io/magmo/apps/c/449.

This is a quick POC of validating transitions by doing a read-only call to the `validTransition` function on the Adjudicator.

I tried this against the `ropsten` network and it seemed to perform reasonably well, I didn't really notice any huge delay making transitions.


lalexgap:
I just realized that `validTransition` is a pure function so we don't even need to use `provider.call` we can just call it.

However the `transactionTester` could be used to verify other transactions in the future if we'd like.

PR #92, opened 2019-01-26
App is responsible for toggling wallet visibility

kerzhner:
With this PR, `displayOutbox` is unused (as far as I can tell). Do we want to keep the wiring for `displayOutbox` around for future features?


kerzhner:
You are totally right @lalexgap. Thanks for the review!

PR #94, opened 2019-01-28
Wallet-client no longer exposes openChannel

kerzhner:
Fixes https://zube.io/magmo/apps/c/476


kerzhner:
@geoknee, yes, I think a compilation sanity check in circle CI is a great idea

PR #100, opened 2019-01-29
Adding ropsten config

kerzhner:


geoknee:
The ropsten config seems good to me, but I wonder if we can fix a bit more in this PR?

I hadn't actually tried to run `yarn build` before. Doing that on this branch gave me two problems: 

1. node ran out of memory when trying to build the app. I fixed this with the following change to `package.json`:
```
  "scripts": {
    "start": "node scripts/start.js",
    "build": "node --max-old-space-size=4076 scripts/build.js",
```
2. node cannot find the contract artifacts:
```
$ node --max-old-space-size=4076 scripts/build.js
Creating an optimized production build...
Failed to compile.

/Users/georgeknee/magmo/apps/packages/tictactoe/src/redux/login/saga.ts
Type error: Cannot find module '../../../build/contracts/TicTacToeGame.json'.  TS2307
```
For this last one, you'll notice that the required folder does not even exist. This is also the case in rps, but does not cause a problem. I believe that rps is doing some funky stuff with resolving paths during the build process, that isn't being replicated correctly in ttt.

When running a `yarn start`, we actually deploy the contracts with truffle, which populates the referenced folder above with the artifacts (so that explains why we see no error there).

geoknee:
Looks like I didn't quite catch what you were driving at @kerzhner. I now understand that by setting environment variables we can deploy to ropsten with `yarn start`. However, it seems we are both getting the same error 
```
index.js:1 uncaught at rootSaga at rootSaga 
 at loginRootSaga 
 at loginStatusWatcherSaga 
 at loginStatusWatcherSaga 
 TypeError: Cannot read property 'address' of undefined
```
Do you think it is a good idea to solve *this* problem in this PR?

 Although I believe what I wrote above is valid, it seems to be not strictly be related to this PR after all.

geoknee:
See https://zube.io/magmo/apps/c/492 and

geoknee:
from @lalexgap : 

I think we also have an issue where running `yarn start` always tries to start up ganache and deploy the contracts there even if using a test net (things still work, it just wastes time spinning up ganache and deploying contracts that never get used)

kerzhner:
@geoknee, an alternative is to deploy contracts to ropsten on every `yarn start` when the app is pointed to ropsten. The downside is that contract deploys can take a while. Either way, contract deployment probably doesn't belong in this PR.

PR #101, opened 2019-01-29
Gk ttt insufficient funds flow

geoknee:
When player A repeatedly beats player B (so that B runs out of funds), the apps enter "Game Over" mode appropriately, and the wallet pops up to help close the channel.


kerzhner:
@geoknee small thing. Looks like `packages/.DS_Store` file was added in this PR. It would be good to keep these macOS files out of the repo.

packages/tictactoe/src/redux/__redux-bookmarks__/notes.md
```diff
@@ -0,0 +1,4 @@
+# Redux bookmarks
```
kerzhner: @geoknee I like this idea, but would other developers be able to reuse these bookmarks? Does redux state contain addresses, requiring your browser + Metamask to hold the correct private keys to sign states + make transactions?

packages/tictactoe/src/redux/game/__tests__/player-a.test.ts
```diff
@@ -10,8 +10,8 @@ import {
 } from "../../../core";
 import * as actions from "../actions";
 import * as state from "../state";
-import BN from "bn.js";
-import bnToHex from "../../../utils/bnToHex";
+// import BN from "bn.js";
```
kerzhner: Any reason not to remove this commented out code? Generally, commented out code reduces readability

packages/tictactoe/src/redux/__redux-bookmarks__/notes.md
geoknee: Yes, I think you're right about this (so have removed). Still a useful workflow, I think, but probably requires each dev to create these bookmarks themselves (once, and not to check them in).

packages/tictactoe/src/redux/game/__tests__/player-a.test.ts
geoknee: 👍

geoknee:
> @geoknee small thing. Looks like `packages/.DS_Store` file was added in this PR. It would be good to keep these macOS files out of the repo.

They are currently gitignored, but I think this one snuck in before that... have removed.

packages/tictactoe/src/redux/__redux-bookmarks__/notes.md
kerzhner: Agreed, I use saved redux state frequently.

PR #104, opened 2019-01-29
Update README

andrewgordstewart:


packages/rps/README.md
```diff
@@ -64,26 +64,28 @@ This will fire up the Storybook panel inside a browser.
     ```
     yarn deployContracts
     ``` 
-    
+
+NOTE: The wallet also has to deploy the simple adjudicator on the same network.
```
lalexgap: I don't think this is true (I was able to deploy `rps` contracts without the wallet contracts deployed.)

geoknee: Is the point that *funding a game* will not work unless wallet and app are pointing to the same network?

packages/rps/README.md
```diff
@@ -64,26 +64,28 @@ This will fire up the Storybook panel inside a browser.
...
+`yarn test:contracts`
 
 #### To run all tests (before submitting a PR):
 
-`yarn run test`
+`yarn test`
 
 #### To update dependencies:
 
 `yarn install`
```
geoknee: Should this be `npx lerna bootstrap` ?

packages/rps/README.md
```diff
@@ -64,26 +64,28 @@ This will fire up the Storybook panel inside a browser.
...
 
 #### To add a dependency:
 
-`yarn add [package-name]` - installs the latest version of the package
+`npx lerna add [package name] --scope=rps --dev` - installs the latest version of the package to the rps package
```
geoknee: Should the `--dev` flag be optional?

packages/rps/README.md
andrewgordstewart: On reflection, what I wrote is not true. I got into a state where the wallet hadn't deployed contracts to the same network, but it wasn't the adjudicator.

packages/rps/README.md
kerzhner:
```suggestion
`npx lerna add [package name] --scope=rps` - installs the latest version of the package to the rps package. Use `--dev` flag to add the new package to `devDependencies` instead of `dependencies`.
```

packages/rps/README.md
kerzhner:
```suggestion
```

PR #108, opened 2019-01-29
Add retry for transaction failure

lalexgap:
Whenever a transaction submission fails (IE: user rejects a transaction) we now catch that and allow then to retry the transaction.

### DIFF
--- a/packages/wallet/src/redux/reducers/funding.ts
+++ b/packages/wallet/src/redux/reducers/funding.ts
@@ -48,9 +48,38 @@ export const fundingReducer = (state: states.FundingState, action: actions.Walle
       return sendFundingDeclinedMessageReducer(state, action);
     case states.ACKNOWLEDGE_FUNDING_DECLINED:
       return acknowledgeFundingDeclinedReducer(state, action);
+    case states.DEPLOY_TRANSACTION_FAILED:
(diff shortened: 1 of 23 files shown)


packages/wallet/src/redux/reducers/funding.ts
```diff
@@ -154,10 +183,7 @@ const aSubmitDeployToMetaMaskReducer = (state: states.ASubmitDeployInMetaMask, a
     case actions.TRANSACTION_SUBMITTED:
       return states.waitForDeployConfirmation({ ...state, transactionHash: action.transactionHash });
     case actions.TRANSACTION_SUBMISSION_FAILED:
-      return states.waitForChannel({
-        ...state,
-        messageOutbox: fundingFailure(state.channelId, "Other", action.error.message),
-      });
+      return states.deployTransactionFailed(state);
```
kerzhner: @lalexgap Checking to see if my understanding is correct here. We are currently not handling the case where transaction submission succeeds, but the transaction is rejected for some reason after the submission.

lalexgap: @kerzhner That's correct. We can probably extend this to cover where a transaction errors out after submission as well. (Although retrying a failed transaction might not help a lot?)

PR #121, opened 2019-01-30
Configurable ganache network id

andrewgordstewart:
The network id used by `startGanache` is now configurable by the command
line or .env, in decreasing order of priority


andrewgordstewart:
I didn't bother with using the known network id in test scenarios because the channel id depends on the latest run of `yarn deployContracts`: [these strings](https://github.com/magmo/apps/blob/master/packages/rps/src/core/test-scenarios.ts#L65-L78
) will no longer be valid (and will change on re-deploy)

PR #127, opened 2019-01-31
Allow challenging in play again

### LINKED ISSUE #125: Cannot challenge after selecting play again
A user cannot challenge after selecting play again. This means if the other user is not responding, they're stuck waiting.

lalexgap:
Fixes #125


kerzhner:
@geoknee we probably want to implement this in tictactoe as well

geoknee:
@kerzhner I think this needs looking at in ttt. I raised a new issue https://github.com/magmo/apps/issues/128 .

PR #131, opened 2019-01-31
Run build as the first command for test:ci

lalexgap:


geoknee:
Looks great, this should inform us about ci failures a bit faster.

PR #150, opened 2019-02-01
Use WithdrawAndClose instead of Close to improve user experience

### LINKED ISSUE #23: Use WithdrawAndClose instead of two separate transactions for Withdraw and Close
We should use the WithdrawAndClose so we only have to use one transaction to end the game instead of the two we use now.

This isn't essential for the wallet to function but it does make it a lot nicer.

lalexgap:
Fixes #23

### DIFF
--- a/packages/wallet/src/components/withdrawing/SelectAddress.tsx
+++ b/packages/wallet/src/components/withdrawing/SelectAddress.tsx
@@ -3,7 +3,7 @@ import React from 'react';
 import { Button } from 'reactstrap';
 import { StyleSheet, css } from 'aphrodite';
 import SidebarLayout from '../SidebarLayout';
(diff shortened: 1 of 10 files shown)


packages/wallet/src/components/withdrawing/SelectAddress.tsx
```diff
@@ -42,25 +46,45 @@ export default class SelectAddress extends React.PureComponent<Props, State> {
...
       </SidebarLayout>
     );
   }
   handleSubmitAddress() {
-    this.props.approveWithdrawal(this.state.withdrawAddress);
+    this.props.approveAction(this.state.withdrawAddress);
   }
 
 }
+const styles = StyleSheet.create({
```
kerzhner: @lalexgap Should this be alongside the rest our sass in stylesheets?

PR #167, opened 2019-02-07
Added basic login error page for when library is not deployed

### LINKED ISSUE #99: Unhandled error when contracts aren't deployed
We probably won't let the deployed app get to this state, but we don't handle the case when fmg-core contracts aren't deployed:
```
index.js:1 uncaught at rootSaga at rootSaga 
 at loginRootSaga 
 at loginStatusWatcherSaga 
 at loginStatusWatcherSaga 
 TypeError: Cannot read property 'address' of undefined
```
([here](https://github.com/magmo/apps/blob/master/packages/rps/src/redux/login/saga.ts#L69))

lalexgap:
Fixes #99.


lalexgap:
> Looks good! I got the page to display by restarting ganache while the app is on the home screen (rps and ttt).
> 
> Could we have the text on the new page centred vertically? Should be easy to copy the pattern from MetamaskErrorPage.

Done, copied the styling from `MetamaskErrorPage`

geoknee:
👍 thanks!

PR #171, opened 2019-02-07
Wait for iFrame to load

### LINKED ISSUE #170: Occasionally the user can get stuck at the login page
Occasionally you can get stuck on the login page and the `Start Playing` button doesn't seem to do anything. 

I suspect this has to do with trying to init the wallet before the iFrame is loaded.

lalexgap:
Fixes #170 .

Ideally I think the `wallet-client` library would wait for the iFrame to load but that would change the `createWalletIframe` function into an `async` function which makes it more awkward to integrate into a react component.

### DIFF
--- a/packages/rps/src/containers/SiteContainer.tsx
+++ b/packages/rps/src/containers/SiteContainer.tsx
@@ -11,12 +11,14 @@ import LoadingPage from '../components/LoadingPage';
 import { createWalletIFrame } from 'wallet-client';
 import { WALLET_IFRAME_ID, WALLET_URL } from '../constants';
 import LoginErrorPage from '../components/LoginErrorPage';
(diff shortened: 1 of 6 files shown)


geoknee:
Perhaps I am doing something wrong, but this actually seemed to make things worse for me, and I was unable to log in.

lalexgap:
> Perhaps I am doing something wrong, but this actually seemed to make things worse for me, and I was unable to log in.

@geoknee Good catch. It was possible that the `LOGIN.WALLET_IFRAME_LOADED` action could fire before the login saga was ready to take it. I've changed it so the first thing we do in the login saga is wait for the `LOGIN.WALLET_IFRAME_LOADED` action.

PR #172, opened 2019-02-08
Gk ttt allow winner to cashout

### LINKED ISSUE #128: Winner should be able to avoid playing again
Either player should be able to resign instead of playing again, should they wish.

Either player should be able to challenge if they are waiting for the other to play again (and the other player has also not resigned).

geoknee:
Fixes #128 .

This PR introduces a new application flow around playing again, which I think better reflects the options available to the player. In particular, it shows a 'waiting' screen to the winner of a round, so that they know they cannot yet resign and must wait for the other player or challenge. If the other player sends a PlayAgainMeFirst position, the winner can then resign without penalty. 

Pending a nicer diagram, this simplified one shows the new flow:

ttt-user-workflow-simplified

The new flow has new screens in storybook, new test-scenarios and tests. I also did a lot of work on the notes.


geoknee:
Fixes #174.

geoknee:
Fixes #178 .

PR #177, opened 2019-02-08
Properly exit case statement if its not message we expect

### LINKED ISSUE #176: Withdraw failing
Repro steps:
1. Play against a bot until you beat the bot, and the bot concludes the game.
2. Attempt to withdraw funds.

lalexgap:
Fixes #176.

The bot was sending a message that we didn't expect and since we lacked the `break` we would cascade into the next case statement and transition to the wrong state.


lalexgap:
> Looks good! Is it worth it to add any unit test(s) for this code path?

Done!

PR #214, opened 2019-02-28
Nitro Wallet Updates

lalexgap:
- Revert funding and tweak to fit nitro adjudicator
- Add work around for adjudicator event tests
- Update to latest `fmg-core` and `nitro-adjudicator`


geoknee:
👍

PR #220, opened 2019-03-04
The client can play against Neo Bot

andrewgordstewart:


packages/wallet/src/redux/reducers/__tests__/closing.test.ts
```diff
@@ -104,7 +105,7 @@ describe('start in WaitForOpponentConclude', () => {
     const fromHexMock = jest.fn().mockReturnValue(concludeCommitment2);
     Object.defineProperty(fmgCore, "fromHex", { value: fromHexMock });
 
-    const action = actions.messageReceived('0x0', '0x0');
+    const action = actions.messageReceived('commitment' as unknown as Commitment, '0x0');
```
lalexgap: This will probably need to be changed to `concludeCommitment2` instead of `'commitment' as unknown as Commitment` to get the test passing (also the `fromHex` mocking can probably go)

packages/rps/src/redux/message-service/saga.ts
```diff
@@ -267,29 +277,58 @@ function* receiveChallengePositionFromWalletSaga() {
...
+}
 
-function* validateMessage(commitment: RPSCommitment, signature) {
+function* validateMessage(commitment: Commitment, signature) {
   try {
-    return yield Wallet.validateCommitmentSignature(WALLET_IFRAME_ID, asCoreCommitment(commitment), signature);
+    return yield Wallet.validateCommitmentSignature(WALLET_IFRAME_ID, commitment, signature);
   } catch (err) {
     if (err.reason === 'WalletBusy') {
+      // Why do you create this event channel? What does 'WalletBusy' really mean?
```
lalexgap: The `wallet` can refuse to sign or validate something if a challenge is ongoing (and possibly some other scenarios?) so  application need to be able to handle the case where the wallet is busy and try again later.
We could rename `WalletBusy` to `ChallengeOngoing` to make this clearer. (maybe add a message as well)

packages/rps/src/redux/open-games/saga.ts
```diff
@@ -1,16 +1,26 @@
-import { fork, take, select, cancel, call, apply } from 'redux-saga/effects';
```
lalexgap: Not really specific to this PR but I think a line width of 80 is a bit too small 😛

packages/rps/src/redux/message-service/saga.ts
andrewgordstewart: 👍

PR #222, opened 2019-03-05
Prettier RPS

andrewgordstewart:


andrewgordstewart:
One can add a prettier write command to a pre-commit hook, in various ways, to make this seamless: https://prettier.io/docs/en/precommit.html

PR #225, opened 2019-03-05
Disambiguate 'message'

andrewgordstewart:
The term 'message' appears overloaded in the context of the wallet-app
relationship:

1. MyApp can "message" MyWallet by calling `Wallet.messageWallet`
2. MyApp can create a "message" for MyWallet by calling any wallet
   function, in wallet-functions, which posts a "message" to the iframe
3. MyWallet can send a message to MyApp by putting a WalletEvent in the wallet's messageOutbox
4. MyApp can message YourApp, where that message could be an
   instruction to message YourWallet (see 1)

The fourth case is the most confusing to me, since the instruction would
then post a message to the wallet, so this action involves three
distinct types of messages.

This PR disambiguates two uses of 'message'
- renaming a "message telling the wallet to do something" to be a instruction.
- renaming "message-to-wallet" as a "wallet-instruction"


packages/magmo-wallet-client/src/wallet-events.ts
```diff
@@ -241,26 +241,48 @@ export const hideWallet = () => ({
...
 
-// MESSAGING
+// WALLET-TO-WALLET COMMUNICATION
 // =========
 /**
- * The type of event when a message to the opponent's wallet is requested.
+ * The type of event when a message relay to the opponent's wallet is requested.
  */
-export const MESSAGE_REQUEST = 'WALLET.MESSAGING.MESSAGE_REQUEST';
+export const MESSAGE_RELAY_REQUESTED = 'WALLET.MESSAGING.MESSAGE_REQUEST';
```
kerzhner: @andrewgordstewart Should the constant value reflect the name?

packages/magmo-wallet-client/src/wallet-events.ts
```diff
@@ -241,26 +241,48 @@ export const hideWallet = () => ({
...
-export const messageRequest = (to: string, data: Commitment | string, signature: string) => ({
-  type: MESSAGE_REQUEST as typeof MESSAGE_REQUEST,
+export const messageRelayRequested = (to: string, data: string) => ({
+  type: MESSAGE_RELAY_REQUESTED as typeof MESSAGE_RELAY_REQUESTED,
   to,
   data,
+});
+
+/**
+ * The event thrown when the wallet requests a message be relayed to the opponent's wallet.
```
kerzhner: Not sure about the "thrown" language here. Typically, "throw" is used for errors and exceptions.

packages/magmo-wallet-client/src/wallet-events.ts
```diff
@@ -241,26 +241,48 @@ export const hideWallet = () => ({
...
+) => ({
+  type: COMMITMENT_RELAY_REQUESTED as typeof COMMITMENT_RELAY_REQUESTED,
+  to,
+  commitment,
   signature,
 });
 
 /**
- * The event thrown when the wallet requests a message be sent to the opponent's wallet.
+ * The event thrown when the wallet requests a commitment be relayed to the opponent's wallet.
```
kerzhner: Same comment about "thrown"

packages/rps/src/redux/message-service/saga.ts
```diff
@@ -32,29 +36,63 @@ export default function* messageSaga() {
...
 
-interface Message {
+interface CommitmentMessage {
   data: Commitment;
   signature: string;
   queue: Queue;
   userName?: string;
 }
 
+interface MessageMessage {
```
kerzhner: Not a big fan of this name. "NonProtocolMessage"? "NonCommitmentMessage"? "MessageForWallet"?

packages/rps/src/redux/message-service/saga.ts
```diff
@@ -32,29 +36,63 @@ export default function* messageSaga() {
...
+  userName?: string;
+}
+
+type Message = CommitmentMessage | MessageMessage;
+
 export function* sendWalletMessageSaga() {
-  const sendMessageChannel = createWalletEventChannel([Wallet.MESSAGE_REQUEST]);
+  const sendMessageChannel = createWalletEventChannel([
+    Wallet.COMMITMENT_RELAY_REQUESTED,
+    MESSAGE_RELAY_REQUESTED,
```
kerzhner:
```suggestion
    Wallet.MESSAGE_RELAY_REQUESTED,
```

packages/rps/src/redux/message-service/saga.ts
```diff
@@ -32,29 +36,63 @@ export default function* messageSaga() {
...
+            signature: theirSignature,
+            type: 'Commitment',
+          });
+        } else {
+          yield call(reduxSagaFirebase.database.create, `/messages/${to.toLowerCase()}`, message);
+        }
+        break;
+      }
+      case Wallet.MESSAGE_RELAY_REQUESTED: {
+        const { data, to } = messageRequest;
```
kerzhner: What about the http communication case?

packages/rps/src/redux/message-service/saga.ts
andrewgordstewart: Also not a fan of this name, which is a motivation for this review. The problem is that this is a message to YourApp containing a message to YourWallet. `NonCommitmentMessage` seems the best out of your suggestions, and I don't have a better one.

kerzhner: Let's go with `NonCommitmentMessage` for now?

PR #233, opened 2019-03-10
Split wallet redux state

andrewgordstewart:
## Motivation
To refactor the wallet state to allow the wallet to manage multiple channels -- to start, a single application channel and a single wallet channel

## Changes
Prior, the entire wallet redux state was at the top level.

Now, it is partitioned into `channelState` and `outboxState`, with some other global attributes at the top level, such as a possible `unhandledAction`.

The wallet state has two stages, `Initializing` and `Initialized`. It never transitions from `Initialized` to `Initializing`.

The channel state is updated with a `channelReducer`. However, sometimes, an update to the channel state necessitates a side effect, so the channel reducer returns a state which looks like a wallet state.

In addition, the private key used by the wallet to sign state updates is no longer static, and a random key is created when the app asks the wallet to initialize an application channel.

The channel state, when it exists, has two stages, `Initializing` (when a commitment is not known, but the wallet has created the user's signing key) and `Initialized`, (when the first commitment is received).


packages/wallet/src/redux/reducers/index.ts
```diff
@@ -1,133 +1,56 @@
...
-  if (conclusionStateFromOwnRequest) {
-    return conclusionStateFromOwnRequest;
-  }
-
-  const conclusionStateFromOpponentRequest = receivedValidOpponentConclusionRequest(state, action);
-  if (conclusionStateFromOpponentRequest) {
-    return conclusionStateFromOpponentRequest;
+    nextOutbox.transactionOutbox = undefined;
   }
+  state = sideEffectsReducer(state, nextOutbox);
```
kerzhner: @andrewgordstewart If I am understanding the logic correctly, when the wallet is in the `CHANNEL_INITIALIZED` state, the `sideEffectsReducer` will get called twice (once in the `walletReducer` and once in `initializedReducer`)? Is this intended?

kerzhner: Also, I think `nextOutbox` might always be an empty object when on this line.

packages/wallet/src/redux/reducers/index.ts
```diff
@@ -1,133 +1,56 @@
...
-    return null;
-  }
-  if (!validTransition(state, commitment)) {
-    return null;
+  if (!sideEffects) {
+    return state;
   }
+  // TODO: We need to think about whether to overwrite existing outbox items.
+  const newState = { ...state, outboxState: { ...state.outboxState } };
+  Object.keys(sideEffects).map(k => (newState.outboxState[k] = sideEffects[k]));
```
kerzhner: A bit confused on what is intended here. `sideEffects` object is changed, but then the object is not used for the rest of the function.

packages/wallet/src/redux/reducers/index.ts
```diff
@@ -1,133 +1,56 @@
...
-  if (!validCommitmentSignature(commitment, signature, opponentAddress)) {
-    return null;
-  }
-  if (!validTransition(state, commitment)) {
-    return null;
+  if (!sideEffects) {
+    return state;
   }
+  // TODO: We need to think about whether to overwrite existing outbox items.
+  const newState = { ...state, outboxState: { ...state.outboxState } };
```
kerzhner: Should we clear the out the `outboxState` on the channel state object?

packages/wallet/src/containers/Challenging.tsx
```diff
@@ -58,7 +58,7 @@ class ChallengingContainer extends PureComponent<Props> {
             Check the progress on&nbsp;
             <EtherscanLink
               transactionID={state.transactionHash}
-              networkId={state.networkId}
+              networkId={-1} // TODO: Fix network id
```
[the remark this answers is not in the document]
andrewgordstewart: ... maybe ;)

packages/wallet/src/redux/reducers/index.ts
andrewgordstewart: I don't understand this question.

packages/wallet/src/containers/Challenging.tsx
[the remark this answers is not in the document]
andrewgordstewart: So, the question here is, should we pass down the network id from the global state, or are we going to move towards a network id as a configuration variable?

packages/wallet/src/redux/states/channels/challenging.ts
```diff
@@ -51,14 +50,12 @@ export interface WaitForChallengeInitiation extends ChannelOpen {
   stage: typeof CHALLENGING;
 }
 export function waitForChallengeInitiation<T extends ChannelOpen>(
```
andrewgordstewart: Is this a mistake?

packages/wallet/src/containers/Challenging.tsx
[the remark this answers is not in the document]
andrewgordstewart: We'll deal with this in a separate PR.

PR #234, opened 2019-03-11
fix storybook by embedding channel states in a wallet state

geoknee:
Otherwise the Wallet container switch is not fed the correct `stage`, and we fall through to the default case every time.

### DIFF
--- a/packages/wallet/src/__stories__/index.tsx
+++ b/packages/wallet/src/__stories__/index.tsx
@@ -1,12 +1,13 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
-import Wallet from '../containers/Wallet';
+import WalletContainer from '../containers/Wallet';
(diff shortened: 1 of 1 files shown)


packages/wallet/src/__stories__/index.tsx
```diff
@@ -62,127 +62,248 @@ const fakeStore = state => ({
...
+  networkId?: number,
+): walletStates.WalletState {
+  return {
+    ...initializedWalletState,
+    channelState: { ...channelState },
+    networkId: networkId || 3,
+  };
+}
+
 const testState = state => () => (
```
andrewgordstewart: Does it make sense to define
```
const testState = channelState => () => (
  <Provider store={fakeStore(walletStateFromChannelState(channelState))}>
    <Wallet />
  </Provider>
);
```

instead?

PR #244, opened 2019-03-13
Factor funding state off of channel state.

andrewgordstewart:
Currently, a channel is always funded directly, by making depositing in to the channel's holdings on the nitro adjudicator.

In the future, a channel might be funded indirectly, via a ledger channel, or virtually, via a guarantor channel.

This PR refactors the channel state, pulling the state related specifically to funding into `walletState.channelState.fundingState`. Currently, the only type of `fundingState` is direct funding, but in the future, an indirect funding state can directly substitute it, in the same slot, and an independent, `indirectFundingReducer` can be used to reduce that part of the funding state.

## TODO:
* [ ] Fix up some of the naming
  - Names will change in planned work
* [x] Figure out which funding tests to keep
* [x] Think a bit more about how the funding reducer should work
* [ ] Make sure that the edge cases previously handled with unhandled actions still work
 - We will refactor the funding states, which will remove edge cases (and maybe add some other ones)


packages/wallet/src/containers/DirectFunding.tsx
```diff
@@ -0,0 +1,104 @@
...
+        );
+
+      // PlayerA
+      case fundingStates.A_WAIT_FOR_DEPOSIT_TO_BE_SENT_TO_METAMASK:
+        return <AFundingStep step={0} />;
+      case fundingStates.A_SUBMIT_DEPOSIT_IN_METAMASK:
+        return <AFundingStep step={1}>Please confirm the transaction in MetaMask!</AFundingStep>;
+      case fundingStates.A_WAIT_FOR_DEPOSIT_CONFIRMATION:
+        return (
+          <AFundingStep step={2}>
```
kerzhner: It looks like this PR did not change this functionality, but leaving this comment anyways... It might be more explicit (and less error prone?) for steps to have names that represent the step functionality as opposed to integer indexes.

packages/wallet/src/redux/states/channels/funding.ts
```diff
@@ -1,235 +1,169 @@
...
-  | AcknowledgeFundingSuccess
+  | BWaitForPostFundSetup
+  | AcknowledgeFundingSuccess // <-- at this point, a message might need to be sent
   | SendFundingDeclinedMessage
-  | AcknowledgeFundingDeclined
-  | ADepositTransactionFailed
-  | BDepositTransactionFailed;
+  | AcknowledgeFundingDeclined;
+
+/**
```
kerzhner: @andrewgordstewart Do we want to keep this comment here?

andrewgordstewart: Probably not!

packages/wallet/src/containers/DirectFunding.tsx
andrewgordstewart: I'll make an issue for this, so we can deal with it in a separate PR.

packages/wallet/notes/desired_redux_diagrams.md
```diff
@@ -0,0 +1,264 @@
+<!-- Recommend VSCode plugin
```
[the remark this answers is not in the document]
andrewgordstewart: Seconding this.

PR #253, opened 2019-03-20
channelState manages multiple channels

andrewgordstewart:
- Defines a `ReducerWithSideEffects` type, which specifies a uniform way in which a reducer should declare side effects
- Defines a `combineReducersWithSideEffects`, which combines reducers-that-have-side-effects, collecting all the side effects along the way
- Changes `channelState` to hold two containers: `initializingChannels` (those for which the wallet only knows the private key) and `initializedChannels`(those for which the channel has received commitment)
- The `initializingChannels` and `initializedChannels` containers have their own reducer
- (Temporary?) the wallet stores the channelId of the application channel, so that it knows what channel to pull out when it receives actions meant for the application channel (those actions often don't have any data other than the type)

### DIFF
--- a/packages/wallet/src/redux/channelState/reducer.ts
+++ b/packages/wallet/src/redux/channelState/reducer.ts
@@ -1,17 +1,4 @@
-import {
-  OPENING,
-  FUNDING,
-  RUNNING,
-  CHALLENGING,
-  RESPONDING,
-  WITHDRAWING,
-  CLOSING,
-  approveConclude,
-  ApproveConclude,
-  acknowledgeConclude,
-  AcknowledgeConclude,
-  ChannelState,
-} from './state';
+import * as states from './state';
 
 import { openingReducer } from './opening/reducer';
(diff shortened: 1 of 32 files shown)


packages/wallet/src/redux/channelState/reducer.ts
```diff
@@ -20,22 +7,105 @@ import { challengingReducer } from './challenging/reducer';
...
+    }
+    const initializingAddresses = new Set(Object.keys(state.initializingChannels));
+    const ourAddress = channel.participants.find(addr => initializingAddresses.has(addr));
+    if (!ourAddress) {
+      return { state };
+    }
+    const ourIndex = channel.participants.indexOf(ourAddress);
+
+    const { address, privateKey } = state.initializingChannels[ourAddress];
+    delete state.initializingChannels[ourAddress];
```
kerzhner: The state object is being modified here. Not sure if that is dangerous given the Redux paradigms.

PR #254, opened 2019-03-21
Fix funding container issue

andrewgordstewart:
This changes the definition of `isDepositing` to consider the case where a `FUNDING_RECEIVED_EVENT` is dispatched which indicates that the channel is funded, before the `TRANSACTION_CONFIRMED` action is dispatched, indicating that the deposit was successful.

edit: I messed up the git history, so putting this on pause for the moment.


tomclose:
Closing for time being

PR #258, opened 2019-03-25
[WIP] Indirect funding

kerzhner:
This PR includes the initial indirect funding state machine. Creating this PR early as other changes to support ledger channel funding will depend on this work.


packages/wallet/src/redux/fundingState/indirectFunding/state.ts
```diff
@@ -0,0 +1,102 @@
+import { SharedIndirectFundingState, INDIRECT_FUNDING } from '../shared/state';
+export { SharedIndirectFundingState, INDIRECT_FUNDING };
+
+// ChannelFundingStatus
+export const FUNDER_CHANNEL_DOES_NOT_EXIST = 'FUNDER_CHANNEL_DOES_NOT_EXIST';
+export const FUNDER_CHANNEL_EXISTS = 'FUNDER_CHANNEL_EXISTS';
+export const CHANNEL_FUNDED = 'CHANNEL_FUNDED';
+
+// Funding status
+export type ChannelFundingStatus =
```
andrewgordstewart: When you define this type like so, it becomes clear that these things are better served being an `enum` rather than a union type.

```
enum Foo {
  bar = 'BAR',
  baz = 'BAZ',
}

interface Bar {
  type: Foo.bar;
}

interface Baz {
  type: Foo.baz;
}

type Foobar = Bar | Baz;

function exhaustPossibilities(foobar: Foobar) {
  switch (foobar.type) {
    case Foo.bar:
      return 'bar';
    case Foo.baz:
      return 'baz';
    default:
      return unreachable(foobar);
  }
}

function missOnePossibility(foobar: Foobar) {
  switch (foobar.type) {
    case Foo.bar:
      return 'bar';
    default:
      return unreachable(foobar); // Argument of type 'Baz' is not assignable to parameter of type 'never'.
  }
}
```

andrewgordstewart: This definition of the `Foobar` type would not work:
```
interface Foobar {
  type: Foo
}
```
The compiler isn't aware that all possibilities are exhausted in the switch block in `exhaustPossibilities`.

But, I think this is ok, since we seem to actually want the following pattern:
- Define a "base state" interface of shared properties
- Create a few states that extend the base state
- Explicitly whitelist valid states as a union type of the defined states

packages/wallet/src/redux/fundingState/indirectFunding/state.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export interface ChannelFunded extends BaseIndirectFundingState {
+  channelFundingStatus: typeof CHANNEL_FUNDED;
+  funderChannelId: string;
+}
+
+export type IndirectFundingState = FunderChannelDoesNotExist | FunderChannelExists | ChannelFunded;
+
+// type guards
+const guardGenerator = <T extends IndirectFundingState>(type) => (
```
andrewgordstewart: `<T extends IndirectFundingState>(type: ChannelFundingStatus)`

packages/wallet/src/redux/fundingState/indirectFunding/state.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export interface BaseIndirectFundingState extends SharedIndirectFundingState {
+  channelFundingStatus: ChannelFundingStatus;
+}
+
+export interface FunderChannelDoesNotExist extends BaseIndirectFundingState {
+  channelFundingStatus: typeof FUNDER_CHANNEL_DOES_NOT_EXIST;
+}
+
+export interface FunderChannelExists extends BaseIndirectFundingState {
```
andrewgordstewart: This might be better as `ChannelNotFunded` (by funder channel), since the funder channel also exists when you're in the `ChannelFunded` state.

packages/wallet/src/redux/internal/actions.ts
```diff
@@ -22,4 +22,23 @@ export const directFundingConfirmed = (channelId: string) => ({
...
 
-export type InternalAction = DirectFundingRequested | DirectFundingConfirmed;
+export const OPEN_CHANNEL_SUCCESS = 'WALLET.INTERNAL.OPEN_CHANNEL_SUCCESS';
+export const openChannelSuccess = (channelId: string) => ({
+  type: OPEN_CHANNEL_SUCCESS as typeof OPEN_CHANNEL_SUCCESS,
+  channelId,
+});
+export type OpenChannelSuccess = ReturnType<typeof openChannelSuccess>;
+
+// TODO: is the last committment needed?
```
andrewgordstewart: `channelId` is ambiguous here: is it the id of the ledger channel, or the channel allocated to once consensus was reached?

This action might need both. But I think it might be pragmatic to include the entire allocation, before and after consensus was reached. That way, an application channel can filter `CONSENSUS_REACHED` events based on whether its funding was changed.

PR #259, opened 2019-03-25
WIP: Ledger Channel Support in Channel Reducer

lalexgap:
This is a WIP of refactoring the current channel reducer and adding support for ledger channels.

// TODO:

- [x] Add Opening states/reducer
- [ ] Add Funding states/reducer
- [ ] Add Closing states/reducer
- [ ] Add Challenging states/reducer
- [ ] Add Responding states/reducer
- [ ] Add tests for the refactored channel reducer
- [ ] Possibly consolidate states.


packages/wallet/src/redux/channelState/challenging/state.ts
```diff
@@ -24,10 +30,11 @@ export function challengeTransactionFailed<T extends ChannelOpen>(
     type: CHALLENGE_TRANSACTION_FAILED,
     stage: CHALLENGING,
     ...channelOpen(params),
+    channelType: 'Application',
```
andrewgordstewart: This should not be a magic constant, but should be defined
`const APPLICATION = 'APPLICATION'`
```
...
channelType: APPLICATION,
```

Or maybe it should even be an enum

packages/wallet/src/redux/channelState/ledger-channel/opening/reducer.ts
```diff
@@ -0,0 +1,190 @@
...
+import * as internalActions from '../../../internal/actions';
+import { bytesFromAppAttributes, AppAttributes } from 'fmg-nitro-adjudicator';
+import { SignedCommitment } from '../../shared/state';
+
+export const openingReducer = (
+  state: channelStates.OpeningState,
+  action: actions.WalletAction,
+): StateWithSideEffects<channelStates.LedgerChannelStatus> => {
+  switch (state.type) {
+    case channelStates.SEND_INITIAL_PRE_FUND_SETUP:
```
andrewgordstewart: It seems to me like there should be a small difference between this reducer and the opening reducer for an application channel:
- An application channel would jump straight to the `WAIT_FOR_PRE_FUND_SETUP` state, since the wallet isn't responsible for sending the prefund setup. (Though maybe the wallet should be)
- If a channel is an application channel, it's assumed that it will be indirectly funded, so it dispatches an `internal.indirectFundingRequested` action

To me, this makes it seem like there doesn't need to be two different reducers. Are you sure there need to be two sets of reducers for each stage?

packages/wallet/src/redux/channelState/ledger-channel/opening/reducer.ts
```diff
@@ -0,0 +1,190 @@
...
+};
+const sendInitialPreFundSetupReducer = (
+  state: channelStates.SendInitialPreFundSetup,
+  action: internalActions.OpenLedgerChannel | actions.WalletAction,
+): StateWithSideEffects<
+  channelStates.WaitForPreFundSetup | channelStates.SendInitialPreFundSetup
+> => {
+  switch (action.type) {
+    case internalActions.OPEN_LEDGER_CHANNEL:
+      const { channelNonce, ourIndex, participants } = state;
```
andrewgordstewart: Why doesn't the state just have a `Channel` property?

packages/wallet/src/redux/channelState/ledger-channel/opening/reducer.ts
```diff
@@ -0,0 +1,190 @@
...
+  opponentCommitment: Commitment,
+  opponentSignature: string,
+  ourIndex: number,
+  participants: string[],
+) => {
+  if (opponentCommitment.commitmentType !== CommitmentType.PreFundSetup) {
+    throw new Error('Expected PrefundSetup commitment.');
+  }
+  if (opponentCommitment.commitmentCount !== 1 - ourIndex) {
+    throw new Error(` Expected commitment count to be ${1 - ourIndex}`);
```
andrewgordstewart: `1` is a magic constant

packages/wallet/src/redux/channelState/challenging/state.ts
lalexgap: This has been obsoleted now but `channelType` is a string enum.

packages/wallet/src/redux/channelState/ledger-channel/opening/reducer.ts
lalexgap: I'm using the existing `FirstCommitmentReceived` state that is shared by all `channel` reducers.

packages/wallet/src/redux/channelState/ledger-channel/opening/reducer.ts
lalexgap: I'm on the fence about this. Right now there's a bunch of small differences that add up and add complexity to just using one reducer: 
 - They listen for different actions: `COMMITMENT_RECEIVED` vs`OWN/OPPONENT_COMMITMENT_RECEIVED`
- They send different messages (`validationSuccess`/`SignatureSuccess` vs `LedgerChannelOpened`)
- Application/Ledger channels will transition to different states in the `Funding` stage. Application channels  transition to `WAIT_FOR_FUNDING_REQUEST`. Ledger channels will need to transition to something like `SEND_FUNDING_REQUEST`
- Ledger channels need to compose their own prefundsetup.
- They start in different states `WAIT_FOR_CHANNEL` for app channels vs `WAIT_FOR_INITIAL_PREFUNDSETUP` or `SEND_INITIAL_PREFUND_SETUP` for ledger channels.
- Ledger channel states need to keep track of the `appChannelId`.

In the case of the `funding` stage the `ledger-channel` and `application-channel` will look quite a bit different since the `ledger-channel` will not be dealing with any user input (`APPROVE_FUNDING`) or waiting for the application to requestFunding.

Also in the `running` stage the `app-channel` is just waiting for the latest commitments exchanged by the application while the `ledger-channel` is waiting for `ALLOCATION_REQUESTS`, exchanging `commitment`s and sending out `CONSENSUS_REACHED`

My main concern is overloading the reducers with too much conditional logic to handle the differences between a `ledger channel` and an `application channel` making the overall code harder to parse.

However all channels follow the same basic pattern of exchanging `prefundsetups`, requesting funding and exchanging `postfundsetups` that would be nice to capture in some kind of shared state/logic.

packages/wallet/src/redux/channelState/challenging/state.ts
andrewgordstewart: ahhh, :+1:

packages/wallet/src/redux/channelState/ledger-channel/opening/reducer.ts
andrewgordstewart: I think a few of those difference are probably worth refactoring anyway:
- the application channel reducer also needs to listen to the same kind of commitments received, through `relayMessage`, when opening/funding a channel
- it might make sense for the wallet to be responsible for the entire pre-app-phase commitments
- the way that the wallet got the opponent's commitment shouldn't affect what kind of action the saga manager dispatches. If there is a need to distinguish who's commitment it is in a `COMMITMENT_RECEIVED` action, then it should probably always be distinguished.
- I think `WAIT_FOR_CHANNEL` and `WAIT_FOR_INITIAL_PREFUND_SETUP` might actually be the same kind of state? I'm not sure.

It seems like there might be a lot of difference in the funding stage, but maybe not in the other stages. I'm not really .sure though, and it might just make sense to have different reducers. I'm also not sure :/

lalexgap:
This is now defunct with our changes to the funding design.

PR #261, opened 2019-03-25
Refactor funding state

andrewgordstewart:
- Actions are organized by domain (some actions are common)
- Some actions are common. To be useful for updating the funding state, these actions now have the channelId on them.
- To support this, `transactionOutbox` items also store the channel id
- The funding state is decomposed into the `indirectFunding` and `directFunding` parts, for the moment.
-  when it receives a `DIRECT_FUNDING_REQUEST`, the `fundingState` reducer creates a new slot in the `directFunding` branch of the funding state
- otherwise, it calls the combinedReducer on the `directFunding` and the (placeholder) `indirectFunding` state


packages/wallet/src/redux/fundingState/actions.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function isfundingAction(action: actions.WalletAction): action is FundingAction {
+  return action.type.match('WALLET.FUNDING') ||
+    actions.internal.isFundingAction(action) ||
+    actions.isCommonAction(action)
+    ? true
+    : false;
+}
+
+// TODO: This is getting large, we should probably split this up into separate types for each stage
```
kerzhner:
```suggestion
```

PR #262, opened 2019-03-26
Change test helpers to check full array

lalexgap:
Previously the test helpers would only check the first item in an outbox array, which meant the helpers could not be used when multiple items are put in an outbox.

`expectSideEffect` has been updated to check at least one item in the array matches the provided `actionOrObject`.


andrewgordstewart:
This is also done in https://github.com/magmo/apps/pull/261

lalexgap:
Handled by #261

PR #263, opened 2019-03-28
Indirect funding proposal

tomclose:
A proposal for how to approach indirect funding.


andrewgordstewart:
So, this proposal seems perfectly reasonable, but it's not really getting at the problem we're all blocked by, which is to somehow get a "coordinator" that can make decisions about state updates based on a global view. This proposal is about how to define the state to enable indirect funding, and not about how to update that state.

Currently, we have
- a state
- and a set of reducers
which has channel state being updated independently from funding state. This is achieved by dispatching an "internal" action.

To me, the most important thing to do right now is to figure out how we could coordinate the existing wallet without using an `actionOutbox`. This means that whoever is working on that problem can, and probably should, forget about indirect funding. The indirect funding state machine might influence the best long-term choice on how to solve the coordination problem, but I think that the direct funding state machine is complicated enough to be instructive.

Therefore, I think we should think about proposals for rewriting the direct-funding process without an `actionOutbox`.

PR #266, opened 2019-03-28
Remove actionOutbox

andrewgordstewart:
To get wallet development started again, we need to figure out how to coordinate state updates in the wallet.

This PR removes the `actionOutbox`, cutting the existing communication between branches of the redux state.

This communication needs to be replaced with a "coordinator" that responds to, eg., a `FUNDING_RECEIVED_EVENT` which completely funds a channel by updating both the funding state (based on the action) and the channel state (based on the updated funding state).

There is one failing test, though to cover the existing wallet requirements, more tests are needed. The initialized reducer needs to be rewritten to make this test pass.

Proposed implementations of the coordinators can be based on this branch.

### DIFF
--- a/packages/wallet/src/redux/__tests__/coordinator.test.ts
+++ b/packages/wallet/src/redux/__tests__/coordinator.test.ts
@@ -56,7 +56,7 @@ const justReceivedPreFundSetupB = {
 };
 
 describe('when a fundingReceivedEvent caused a channel to be funded', () => {
-  it('updates the corresponding channel state', () => {
+  it.skip('updates the corresponding channel state', () => {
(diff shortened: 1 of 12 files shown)


andrewgordstewart:
The current change to the `initialized` reducer is biased towards a "update state and then coordinate" pattern. I think this is the least risky way to proceed, in terms of starting to make progress again, and rewriting as little code as possible. It might be the wrong choice long-term, but that is fine if at least one of these is true:
- there is little effort required in writing the coordinator to make the tests pass
- writing the coordinator in this way is an intermediate step towards the "right" long-term choice

PR #267, opened 2019-03-28
Proposed coordinator

andrewgordstewart:


packages/wallet/src/redux/initialized/reducer.ts
```diff
@@ -28,6 +32,39 @@ const combinedReducer = combineReducersWithSideEffects({
...
+  const { channelId } = action;
+  const channelStatus = getChannelStatus(state, channelId);
+  if (channelStatus.type !== WAIT_FOR_FUNDING_AND_POST_FUND_SETUP) {
+    return state;
+  }
+  const fundingStatus = getDirectFundingStatus(state, channelId);
+
+  const newState = { ...state };
+  if (stateIsChannelFunded(fundingStatus)) {
+    const { state: newChannelStatus, sideEffects } = fundingReducer(
```
andrewgordstewart: This might be an explicit transition, in which case we can completely get rid of "internal actions"

kerzhner:
@andrewgordstewart Do you think we still need this PR?

PR #268, opened 2019-03-29
Simplify top-level wallet state

tomclose:
This PR removes the `initializing` and `initialized` stages, replacing them with a top-level wallet state with four different types.

The motivation here was to try and simplify things a bit, in preparation for a potential move to actions-first reducers.


packages/wallet/src/redux/state.ts
```diff
@@ -1,10 +1,85 @@
...
+// -----------
+export const WAIT_FOR_LOGIN = 'INITIALIZING.WAIT_FOR_LOGIN';
+export const METAMASK_ERROR = 'INITIALIZING.METAMASK_ERROR';
+export const WAIT_FOR_ADJUDICATOR = 'INITIALIZING.WAIT_FOR_ADJUDICATOR';
+export const WALLET_INITIALIZED = 'WALLET.INITIALIZED';
 
-export type WalletState = InitializingState | InitializedState;
+// ------
+// States
+// ------
```
tomclose: @andrewgordstewart do you think we can safely remove `channelState`, `fundingState` and `outboxState` properties from everything but the `Initialized` state?

andrewgordstewart: Yes

packages/wallet/src/redux/state.ts
```diff
@@ -1,10 +1,85 @@
...
+  return { type: METAMASK_ERROR, outboxState, fundingState, channelState };
+}
+
+export function waitForAdjudicator(params: Properties<WaitForAdjudicator>): WaitForAdjudicator {
+  const { outboxState, fundingState, channelState, uid } = params;
+  return { type: WAIT_FOR_ADJUDICATOR, outboxState, fundingState, channelState, uid };
+}
+
+export function initialized(params: Properties<Initialized>): Initialized {
+  const { outboxState, fundingState, channelState, uid, networkId, adjudicator } = params;
```
lalexgap: Minor but is there a reason we can't just do this?:
 ```
return {
    type: WALLET_INITIALIZED,
    ...params
}
```

tomclose: Yeah - `Properties<Initialized>` could also contain extra keys. (I'm not sure if that's actually important here, but it was useful when we were constructing channel states.)

PR #272, opened 2019-04-01
Remove relayCommitment

andrewgordstewart:
In preparation for messages that belong to a particular process, this PR removes the `relayCommitment` wallet function.

In the wallet, the `message-listener` peeks at the shape of the `data` and dispatches either `messageReceived` or `commitmentReceived`.

The next step would be to place an identifier on the message to indicate that it belongs to the `indirect-funding` "process" or "module", whatever that gets called.


kerzhner:
@andrewgordstewart Perhaps I am out of the loop, but why do we need the differentiation between `messageReceived` and `commitmentReceived`? What about moving to a model where every message to the wallet contains a required `process` field and an optional `commitment` field?

PR #274, opened 2019-04-01
Add a procedure to Transaction actions and outbox

lalexgap:
Added a `procedure` property to all `Transaction` actions and `transactionOutbox`.

Added `todo` comments where the `transactionOutbox` is being set in the `channelReducer`. We'll eventually need to refactor our `challenging`/`closing`/`withdrawing` channel reducers to no longer send a transaction and handle this in `directFundingReducer` or `indirectFundingReducer`

### DIFF
--- a/packages/magmo-wallet-client/src/wallet-types.ts
+++ b/packages/magmo-wallet-client/src/wallet-types.ts
@@ -1,1 +1,1 @@
-export type WalletProcess = 'indirect_funding' | 'direct_funding';
+export type WalletProcess = 'IndirectFunding' | 'DirectFunding';
--- a/packages/magmo-wallet-client/src/wallet-types.ts
+++ b/packages/magmo-wallet-client/src/wallet-types.ts
@@ -1,1 +1,0 @@
-export type WalletProcess = 'IndirectFunding' | 'DirectFunding';
--- a/packages/magmo-wallet-client/src/wallet-types.ts
+++ b/packages/magmo-wallet-client/src/wallet-types.ts
@@ -1,0 +1,1 @@
+export type WalletProcess = 'indirect_funding' | 'direct_funding';
--- a/packages/wallet/src/contract-tests/transactions.test.ts
+++ b/packages/wallet/src/contract-tests/transactions.test.ts
@@ -7,6 +7,7 @@ import {
   transactionFinalized,
   transactionSentToMetamask,
   transactionSubmitted,
+  WalletProcess,
 } from '../redux/actions';
 import { transactionSender } from '../redux/sagas/transaction-sender';
 import { signCommitment, signVerificationData } from '../utils/signing-utils';
@@ -47,7 +48,7 @@ describe('transactions', () => {
   }
 
   async function testTransactionSender(transactionToSend) {
-    const process = 'DirectFunding';
+    const process = WalletProcess.DirectFunding;
     const channelId = 'channelId';
     const saga = transactionSender(transactionToSend, channelId, process);
     saga.next();
--- a/packages/wallet/src/contract-tests/transactions.test.ts
+++ b/packages/wallet/src/contract-tests/transactions.test.ts
@@ -7,7 +7,6 @@ import {
   transactionFinalized,
   transactionSentToMetamask,
   transactionSubmitted,
-  WalletProcedure,
 } from '../redux/actions';
 import { transactionSender } from '../redux/sagas/transaction-sender';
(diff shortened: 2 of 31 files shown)


packages/wallet/src/contract-tests/transactions.test.ts
```diff
@@ -47,9 +47,11 @@ describe('transactions', () => {
   }
 
   async function testTransactionSender(transactionToSend) {
-    const saga = transactionSender(transactionToSend, 'channelId');
+    const process = 'DirectFunding';
```
andrewgordstewart: I dislike the word `process`, because you are overwriting a global variable here.

packages/wallet/src/redux/channelState/challenging/__tests__/challenging.test.ts
```diff
@@ -70,7 +70,7 @@ describe('when in INITIATE_CHALLENGE', () => {
   const state = states.waitForChallengeInitiation(defaults);
 
   describe('when a challenge is initiated', () => {
-    const action = actions.transactionSentToMetamask(channelId);
+    const action = actions.transactionSentToMetamask(channelId, 'DirectFunding');
```
andrewgordstewart: I'd prefer this to be a named constant.

packages/wallet/src/redux/channelState/challenging/reducer.ts
```diff
@@ -64,7 +64,14 @@ const challengeTransactionFailedReducer = (
       );
       return {
         state: states.waitForChallengeInitiation(state),
-        sideEffects: { transactionOutbox: { transactionRequest, channelId: state.channelId } },
+        // TODO: This will be factored out as channel reducers should not be sending transactions itself
+        sideEffects: {
+          transactionOutbox: {
+            transactionRequest,
+            channelId: state.channelId,
+            process: 'DirectFunding',
```
andrewgordstewart: (Name this too)

packages/magmo-wallet-client/src/wallet-types.ts
```diff
@@ -0,0 +1 @@
+export type WalletProcess = 'IndirectFunding' | 'DirectFunding';
```
andrewgordstewart: As per our discussion, this probably belongs in the `wallet` package.

packages/wallet/src/redux/channelState/challenging/__tests__/challenging.test.ts
lalexgap: This is obsolete now but it was typed to a `WalletProcess` type so it would only accept 'DirectFunding` or `IndirectFunding`

packages/wallet/src/redux/channelState/challenging/reducer.ts
lalexgap: See above.

packages/wallet/src/redux/channelState/challenging/__tests__/challenging.test.ts
andrewgordstewart: I agree, but if you want to change `DirectFunding` to something else, currently you have to replace-in-all-files and hope you don't hit a false positive.

If it is a constant that is exported, you rename the constant and you're good.

packages/wallet/src/contract-tests/transactions.test.ts
lalexgap: Renamed to procedure

packages/wallet/src/redux/channelState/challenging/__tests__/challenging.test.ts
lalexgap: I've changed `WalletProcedure` to a string enum.

packages/wallet/src/redux/types.ts
```diff
@@ -0,0 +1,4 @@
+export const enum WalletProcedure {
```
lalexgap: This might not be the best place for this enum but I think it's better than being in the actions file. We can move as needed.

packages/magmo-wallet-client/src/wallet-types.ts
lalexgap: This has been moved to a file in the wallet and `magmo-wallet-client` passes procedure as as string

packages/wallet/src/redux/channelState/challenging/reducer.ts
lalexgap: Not applicable with the string enum changes.

Issue #270, opened 2019-04-01
TransactionSender must accept and dispatch actions with the Process and ChannelId


lalexgap:
- `TransactionOutboxItem` must include `process` and `channelId` properties.
- Any actions the `TransactionSender` dispatches include a `process` and `channelId` properties.

PR #279, opened 2019-04-02
Update Wallet Messages to have a process and channelId

lalexgap:
This is based on Andrew's work in  #272  and my transaction  work in #270 .
- There is no longer separate methods for communicating commitments and messages to a wallet. There is only `relayWalletMessage` and `messageReceived`.
- Wallet messages now include `channelId` and `procedure`.
- Wallet messages are encapsulated in a `WalletMessage` object so the application can blindly relay the message without knowing the contents.
- `RPS` has been updated to the latest wallet messaging changes.

### DIFF
--- a/packages/rps/src/redux/message-service/saga.ts
+++ b/packages/rps/src/redux/message-service/saga.ts
@@ -40,7 +40,7 @@ interface AppMessage {
 }
 
 interface WalletMessage {
-  message: any;
+  payload: any;
   queue: Queue.WALLET;
   userName?: string;
 }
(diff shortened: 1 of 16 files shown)


packages/rps/src/redux/message-service/saga.ts
```diff
@@ -27,87 +28,48 @@ export default function* messageSaga() {
...
+  commitment: Commitment;
   signature: string;
-  queue: Queue;
+  queue: Queue.GAME_ENGINE;
   userName?: string;
 }
 
-interface NonCommitmentMessage {
-  data: string;
+interface WalletMessage {
```
tomclose: It's potentially confusing that this is a different `WalletMessage` from the one defined above. Maybe the other one should be `WalletMessagePayload` or similar.

lalexgap: Good point, done!

PR #287, opened 2019-04-02
channelState funding reducer for both ledger and application channels 

kerzhner:
The vision is that `channelState` funding reducer is shared between application and ledger channels. At the moment, wallet display logic is part of the `FUNDING` stage. The proposal is that the `FUNDING` stage will no longer be responsible for display logic. The funding `procedure` (direct or indirect) will decide when user approval or action is needed **before** the funding stage is entered. Also, currently, user action is requested for acknowledgment of funding success or failure. Once again, the funding `procedure` will handle this.

At the moment, `DIRECT_FUNDING_CONFIRMED` action is part of the `FUNDING` stage reducer. The `FUNDING` stage will be agnostic of the type of funding that funds a channel.

### DIFF
--- a/packages/wallet/src/__stories__/index.tsx
+++ b/packages/wallet/src/__stories__/index.tsx
@@ -89,7 +89,9 @@ const initializedWalletState = walletStates.initialized({
   unhandledAction: undefined,
   outboxState: EMPTY_OUTBOX_STATE,
   channelState: {
-    initializedChannels: { [channelId]: channelStates.approveFunding({ ...playerADefaults }) },
+    initializedChannels: {
+      [channelId]: channelStates.waitForFundingRequest({ ...playerADefaults }),
+    },
     initializingChannels: {},
   },
   networkId: 4,
@@ -164,7 +166,6 @@ function addStoriesFromCollection(collection, chapter, renderer = channelStateRe
 }
 
 const WalletScreensFundingPlayerA = {
-  ApproveFunding: { channelState: channelStates.approveFunding(playerADefaults) },
   WaitForTransactionSent: {
     channelState: channelStates.waitForFundingAndPostFundSetup(playerADefaults),
     fundingState: fundingStates.depositing.waitForTransactionSent(defaultFundingState),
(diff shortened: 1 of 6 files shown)


packages/wallet/src/redux/channelState/funding/reducer.ts
```diff
@@ -19,7 +12,7 @@ import { StateWithSideEffects } from '../../utils';
 
 export const fundingReducer = (
   state: states.FundingState,
-  action: actions.ChannelAction | internal.DirectFundingConfirmed,
+  action: actions.ChannelAction | internal.fundingConfirmed,
```
andrewgordstewart: `internal.FundingConfirmed` ?

kerzhner: Yep, you are right, good call.

packages/wallet/src/redux/channelState/funding/reducer.ts
```diff
@@ -41,8 +31,6 @@ export const fundingReducer = (
     // Setup funding process
     case states.WAIT_FOR_FUNDING_REQUEST:
```
andrewgordstewart: Consider that the `indirectFundingReducer` will need to react to an action to get into the `WAIT_FOR_APPROVAL` state. What action could that be? I think the only sensible action is `FUNDING_REQUESTED`.

Therefore, you might also want to remove this state, leaving the channel status in the `WAIT_FOR_FUNDING_AND_POST_FUND_SETUP` stage and leaving it up to the funding-related code to deal with this.

packages/wallet/src/__stories__/index.tsx
```diff
@@ -89,7 +89,9 @@ const initializedWalletState = walletStates.initialized({
   unhandledAction: undefined,
   outboxState: EMPTY_OUTBOX_STATE,
   channelState: {
-    initializedChannels: { [channelId]: channelStates.approveFunding({ ...playerADefaults }) },
+    initializedChannels: {
+      [channelId]: channelStates.waitForFundingRequest({ ...playerADefaults }),
```
andrewgordstewart: (See comment below)

tomclose:
Wanted to get this merged to build on top of it. Andrew's comments are additions (or additional removals), so can (and should) be addressed in a later PR.

packages/wallet/src/redux/channelState/funding/reducer.ts
kerzhner: I agree with this thinking. https://github.com/magmo/apps/pull/301

PR #288, opened 2019-04-02
Add indirect-funding actions

andrewgordstewart:
Adds some actions that will be routed to the `indirectFundingReducer`


packages/wallet/src/redux/indirect-funding/player-a/actions.ts
```diff
@@ -0,0 +1,26 @@
+import { WalletProcedure } from '../../types';
```
kerzhner: What are your thoughts on file and directory naming conventions? At the moment, we are snake casing file and directory names: `channelState`, `directFunding`, etc

packages/wallet/src/redux/indirect-funding/player-a/actions.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Commitment } from 'fmg-core/lib/commitment';
+
+export const STRATEGY_APPROVED = 'WALLET.INDIRECT_FUNDING.STRATEGY_APPROVED';
+export const strategyApproved = (
+  channelId: string,
+  procedure: WalletProcedure.IndirectFunding,
+) => ({
+  type: STRATEGY_APPROVED as typeof STRATEGY_APPROVED,
+  channelId,
+  procedure,
```
kerzhner: Does this need to be a parameter? Is this action used for any other procedure?

andrewgordstewart: an action without a procedure can't be routed -- that's the point of the `procedure` property, isn't it?

packages/wallet/src/redux/indirect-funding/player-a/actions.ts
andrewgordstewart: I think you mean camel casing. I guess I prefer camel casing, but was matching the directory names from this PR: https://github.com/magmo/apps/pull/285

I think I'll change them here, and request changes in #285

packages/wallet/src/redux/indirect-funding/player-a/actions.ts
kerzhner: Never mind, I see what is happening here.

packages/wallet/src/redux/indirect-funding/player-a/actions.ts
kerzhner: 👍

PR #290, opened 2019-04-02
Indirect Funding Player A states

### LINKED ISSUE #280: Player A states

lalexgap:
Added player A states for `indirect-funding`. This assumes that there is no strategy negotiation. 

closes #280


packages/wallet/src/redux/funding/indirect-funding/player-a/state.ts
```diff
@@ -0,0 +1,76 @@
+import { Properties } from '../../../utils';
```
kerzhner: Commented on Andrew's PR as well about this, but we should be careful about directory and file name convention. In other parts of our code, we are currently using camel case like `channelState` as opposed to dashes like `indirect-funding`

packages/wallet/src/redux/funding/indirect-funding/player-a/state.ts
```diff
@@ -0,0 +1,76 @@
...
+import { PlayerIndex } from '../../../types';
+
+export const WAIT_FOR_APPROVAL = 'WAIT_FOR_APPROVAL';
+export const WAIT_FOR_PRE_FUND_SETUP_1 = 'WAIT_FOR_PRE_FUND_SETUP_1';
```
kerzhner: Should this be `WAIT_FOR_PRE_FUND_SETUP`? The fact that it is the first prefund state is implied by the `player` field.

andrewgordstewart: I kind of agree with @kerzhner here, but after that change, it starts to seem silly to have separate states for player A and player B -- they share exactly the same code, except they have a different value in the `playerIndex` property.

packages/wallet/src/redux/indirectFunding/playerB/state.ts
```diff
@@ -0,0 +1,76 @@
+import { Properties } from '../../utils';
+import { PlayerIndex } from '../../types';
+
+export const WAIT_FOR_APPROVAL = 'WAIT_FOR_APPROVAL';
+export const WAIT_FOR_PRE_FUND_SETUP_0 = 'WAIT_FOR_PRE_FUND_SETUP_0';
+export const WAIT_FOR_DIRECT_FUNDING = 'WAIT_FOR_DIRECT_FUNDING';
+export const WAIT_FOR_POST_FUND_SETUP_0 = 'WAIT_FOR_POST_FUND_SETUP_0';
+export const WAI_FOR_LEDGER_UPDATE_0 = 'WAI_FOR_LEDGER_UPDATE_0';
```
andrewgordstewart: Typo

packages/wallet/src/redux/indirectFunding/playerA/state.ts
```diff
@@ -0,0 +1,76 @@
+import { Properties } from '../../utils';
+import { PlayerIndex } from '../../types';
+
+export const WAIT_FOR_APPROVAL = 'WAIT_FOR_APPROVAL';
+export const WAIT_FOR_PRE_FUND_SETUP_1 = 'WAIT_FOR_PRE_FUND_SETUP_1';
+export const WAIT_FOR_DIRECT_FUNDING = 'WAIT_FOR_DIRECT_FUNDING';
+export const WAIT_FOR_POST_FUND_SETUP_1 = 'WAIT_FOR_POST_FUND_SETUP_1';
+export const WAI_FOR_LEDGER_UPDATE_1 = 'WAI_FOR_LEDGER_UPDATE_1';
```
andrewgordstewart: typo

packages/wallet/src/redux/funding/indirect-funding/player-a/state.ts
andrewgordstewart: We're going to stick with the code as is. It's more complicated than it needs to be, but it's already written, and tom seems to want it to be explicit.

PR #299, opened 2019-04-03
kebab-case for file and folder names

tomclose:
Change folder names to ~snake-case~ kebab-case to be consistent with the rest of the project (e.g. `contract-tests`, `magmo-client-wallet`, `message-service` etc.).

Propose that the only time we don't use kebab-case is when the file contains a single export that will be called capitalized as CamelCase. Main examples here are components and containers e.g. `ApplicationContainer.tsx`.


tomclose:
(Though I'm also up for making everything snake-case.)

geoknee:
I think this convention is actually called "kebab-case". 

snake_case_is_like_this 

😉

PR #303, opened 2019-04-03
indirectFundingReducer tests

andrewgordstewart:


packages/wallet/src/redux/indirect-funding/reducer.ts
```diff
@@ -32,7 +32,14 @@ function playerBReducer(
   state: states.Initialized,
   action: actions.indirectFunding.Action,
 ): states.Initialized {
-  return state;
+  return {
+    ...state,
+    indirectFunding: states.indirectFunding.playerB.waitForDirectFunding({
+      ledgerId: 'ledgerId',
```
kerzhner: Maybe leave a TODO that this hardcoding get's removed? Or maybe a comment with a Zube card reference?

andrewgordstewart: I won't bother, because the `playerBReducer` is simply a placeholder to disappear when we write the actual playerB reducer

kerzhner: 👍

PR #311, opened 2019-04-05
Route actions in the channel reducer based on action channelId/commitment

### LINKED ISSUE #310: Channel State Reducer needs to update state for non-app channels
Currently the channel state reducer updates individual channel state based on the current `appChannelId` stored in state.

We need to allow the channel state reducer to handle actions for ledger channels as well.

To handle this we can route a channel action based on its `channelID` or `Commitment`. If neither is present we can defer to `appChannelId` as we do now.

lalexgap:
Allows for handling actions for ledger Channel which would of been ignored based on the `activeAppChannelId`

Fixes #310

### DIFF
--- a/packages/wallet/src/redux/channel-state/reducer.ts
+++ b/packages/wallet/src/redux/channel-state/reducer.ts
@@ -98,9 +98,16 @@ const initializedChannels: ReducerWithSideEffects<states.InitializedChannelState
   if (action.type === actions.CHANNEL_INITIALIZED) {
     return { state };
   }
-  const { appChannelId } = data;
 
-  const existingChannel = state[appChannelId];
+  // If an action has a channelId/commitment we update the channel state for that channel
+  let channelId = data.appChannelId;
+  if ('channelId' in action) {
+    channelId = action.channelId;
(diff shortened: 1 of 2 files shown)


packages/wallet/src/redux/channel-state/reducer.ts
```diff
@@ -111,10 +118,10 @@ const initializedChannels: ReducerWithSideEffects<states.InitializedChannelState
     action,
   );
 
-  return { state: { ...state, [appChannelId]: newState }, sideEffects: outboxState };
+  return { state: { ...state, [channelId]: newState }, sideEffects: outboxState };
 };
 
-const initializedChannelStatusReducer: ReducerWithSideEffects<states.ChannelStatus> = (
+export const initializedChannelStatusReducer: ReducerWithSideEffects<states.ChannelStatus> = (
```
lalexgap: This is a bit ugly but exporting this allows us to mock it out.
It makes the routing tests a lot less fragile since they don't have to rely on logic of the channel stage reducers. 

Perhaps these should just be separated out into their own file?

PR #314, opened 2019-04-05
Route indirectFunding actions to indirectFundingReducer

tomclose:


tomclose:
Closing this as we're now using `processId` so will need to rewrite this anyhow.

PR #317, opened 2019-04-07
Remove WaitForChannel

tomclose:
The `WaitForChannel` state (storing just our `address` and `privateKey` in a channel state) doesn't make sense now that we have the `InitializingChannels` store. This PR removes it.


packages/wallet/src/redux/channel-state/reducer.ts
```diff
@@ -84,12 +63,166 @@ const initializingChannels: ReducerWithSideEffects<states.InitializingChannelSta
...
     },
     sideEffects: { messageOutbox: channelInitializationSuccess(wallet.address) },
   };
 };
 
+type CommitmentReceived = actions.OwnCommitmentReceived | actions.OpponentCommitmentReceived;
+const handleFirstCommmit = (
+  state: states.ChannelState,
+  action: CommitmentReceived,
+): StateWithSideEffects<states.ChannelState> => {
```
lalexgap: This function is pretty long so it might be nice to split it into two, something like `initializeChannel` and `handleCommitment`?

tomclose: I think I will leave this for a future refactor. I'm going to be doing some work on the channel state reducer as part of the challenge work.

PR #319, opened 2019-04-08
Try out repetition over abstraction

tomclose:
I wanted to see whether embracing repetition would make the code clearer in this case. For me, it makes it easier to understand the structure of the states at a glance. I realise there are tradeoffs here and if there were many more states it might get cumbersome. Interested to hear people's thoughts.

### DIFF
--- a/packages/wallet/src/redux/indirect-funding/player-a/state.ts
+++ b/packages/wallet/src/redux/indirect-funding/player-a/state.ts
@@ -7,31 +7,36 @@ export const WAIT_FOR_DIRECT_FUNDING = 'WAIT_FOR_DIRECT_FUNDING';
 export const WAIT_FOR_POST_FUND_SETUP_1 = 'WAIT_FOR_POST_FUND_SETUP_1';
 export const WAIT_FOR_LEDGER_UPDATE_1 = 'WAIT_FOR_LEDGER_UPDATE_1';
 
-interface BasePlayerAState {
+export interface WaitForApproval {
+  type: typeof WAIT_FOR_APPROVAL;
   channelId: string;
   player: PlayerIndex.A;
 }
 
-interface LedgerChannelExists extends BasePlayerAState {
+export interface WaitForPreFundSetup1 {
+  type: typeof WAIT_FOR_PRE_FUND_SETUP_1;
+  channelId: string;
+  player: PlayerIndex.A;
   ledgerId: string;
 }
 
-export interface WaitForApproval extends BasePlayerAState {
-  type: typeof WAIT_FOR_APPROVAL;
-}
-
-export interface WaitForPreFundSetup1 extends LedgerChannelExists {
-  type: typeof WAIT_FOR_PRE_FUND_SETUP_1;
-}
-
-export interface WaitForDirectFunding extends LedgerChannelExists {
+export interface WaitForDirectFunding {
   type: typeof WAIT_FOR_DIRECT_FUNDING;
+  channelId: string;
+  player: PlayerIndex.A;
+  ledgerId: string;
 }
-export interface WaitForPostFundSetup1 extends LedgerChannelExists {
+export interface WaitForPostFundSetup1 {
   type: typeof WAIT_FOR_POST_FUND_SETUP_1;
+  channelId: string;
+  player: PlayerIndex.A;
+  ledgerId: string;
 }
-export interface WaitForLedgerUpdate1 extends LedgerChannelExists {
+export interface WaitForLedgerUpdate1 {
   type: typeof WAIT_FOR_LEDGER_UPDATE_1;
+  channelId: string;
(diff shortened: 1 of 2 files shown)


lalexgap:
I think having all the properties defined on a state does make it easier to understand. In this case the states seems simple enough that either approach is pretty understandable.

 In general my only concern would be that we'll end up with states using slightly different names for properties (IE: `LedgerId`, `LedgerChannelId`) which could happen when we have states defined in multiple files. I guess that's pretty easy to avoid though.

tomclose:
The type system is also good at catching those sorts of errors.

andrewgordstewart:
This definitely feels wrong to me.

Extending an interface means agreeing to a contract. It's either the case like here (A), where it's easy to understand what `LedgerChannelExists` means, or you're in a world (B) where it's hard to understand what `LedgerChannelExists` means.

## Benefits of abstraction
### 1
In case (A), the benefit is low, since it's easy to keep in your head that, for instance, `LedgerChannelExists` means you have three properties.

In case (B), repeating the code makes it hard to understand if you're upholding the intended contract. How does the type system help, when you've defined `SomeType = TypeA | TypeB | TypeC | TypeD`, and the various types  don't quite have the right properties? I can definitely see some cases where the type checker will help you out, but I don't understand _exactly_ what the type checker will complain about in the same way as I do when `TypeX` extends a common interface.

### 2
It's easier to refactor.  In your proposal, if you want to change the variable `ledgerId`, let's say using VScode, you need to `fn+f2` in seven different places. This is on a small set of states.

### 3
Extending a common interface serves as documentation: in the context of this changeset, for instance:
- Before, I understand that every state needs to have a `player` and a `channelId`
- After, I need to
  * look at the definition of the union type to see what states there are
  * cycle through each state
  * keep track of what properties are shared so far

## What if I just want to see what a `WaitForPreFundSetup1` state looks like?

If you're interested to know exactly what the `WaitForPreFundSetup1` interface means, in one place, the type checker is happy to tell you, with a little wrangling:
```
interface Foo {
  foo: 'foo'
}
interface Bar {
  bar?: 'bar'
}
interface FooBar extends Foo, Bar {
  baz: any
}

type T = keyof FooBar
// VSCode tells you:
// type T = "baz" | "foo" | "bar"

const a: FooBar = {}
// VSCode tells you:
// const a: FooBar
// Type '{}' is missing the following properties from type 'FooBar': baz, foo ts(2739)
```
I wonder if there is a more elegant way to do this -- I couldn't find one. Note: the first way fails to inform you about optional properties, and the second fails to tell you that the property is optional. Incidentally, the two techniques combined do tell you which properties are optional.

I wouldn't want to do this regularly, but I believe that the number of times when you need to do this is low enough that the benefit of having the properties explicitly written in the source code is not high enough, compared to the costs of having to write all the properties everywhere needed, in the source code.

tomclose:
@andrewgordstewart to be clear, I'm not suggesting we necessarily do this everywhere. The question I'm asking is whether in this specific case (where we have 4 simple states that easily fit on a single screen), whether it's worth using extension (and introducing two extra concepts into the code).

I'm also not suggesting that the previous way isn't workable. As you highlight, you can always use VSCode to find out what the properties on these objects are.

What I am suggesting is that the repetition way is _better_ in this case. And when I say better, I'm working from the belief that the way a given piece of code is written should be optimized to be read and understood by others.

The reason I believe it is better in this case is that we only have four states and it is easy to write them out (and, if you change a property, easy to make the change). The way it is currently written, in order to understand the shape of `WaitForDirectFunding` (when reading the code on github for example) I have to go on a hunting expedition: I need to locate `LedgerChannelExists` and then `BasePlayerAState` and put all the properties together in my head. If I'm in VSCode I can use the editor to help, but I need to write something down to trick it into telling me. If we write them out, I can immediately see the shape and get a better understanding of what's going on. Human brains are good at spotting repetition and deviations from repetition. I'm arguing that in seeing those four states on the screen I immediately get a picture of how the structure of the different states and how they relate to one another, that is clearer than if I have to construct them in my head.

Overall this case itself is not particularly significant but I believe the principle is. In particular, it's the notion that "this definitely feels wrong" that I want to challenge. I believe that we should always try to make the code that is written as easy to understand as possible and sometimes repetition is a useful tool for doing this.

tomclose:
> How does the type system help, when you've defined SomeType = TypeA | TypeB | TypeC | TypeD, and the various types don't quite have the right properties?

What I meant here is that if you then try to use one of the properties you expect to be on `SomeType` but that isn't, the type checker will complain. E.g if you missed defining a property on one of them or misspelt it or something.

andrewgordstewart:
I had not considered readability on Github, and I would agree that in this case, it's more clear in this PR!

tomclose:
Just going to close this because it's stale.

PR #320, opened 2019-04-08
Deduce playerA from state type

tomclose:
This is an alternative to putting the `player` explicitly on the state. It means that the transitions become fully defined by the `type` alone instead of the `(player, type)` pair, which seems like a nice property. (The developer does have to be careful to avoid name collisions on types though, but that's already the case.)

Interested to hear thoughts on this.

### DIFF
--- a/packages/wallet/src/redux/indirect-funding/player-a/state.ts
+++ b/packages/wallet/src/redux/indirect-funding/player-a/state.ts
@@ -1,41 +1,36 @@
 import { Properties } from '../../utils';
-import { PlayerIndex } from '../../types';
+import { IndirectFundingState } from '../state';
 
-export const WAIT_FOR_APPROVAL = 'WAIT_FOR_APPROVAL';
+export const WAIT_FOR_APPROVAL = 'WAIT_FOR_APPROVAL_A';
 export const WAIT_FOR_PRE_FUND_SETUP_1 = 'WAIT_FOR_PRE_FUND_SETUP_1';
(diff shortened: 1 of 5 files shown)


packages/wallet/src/redux/indirect-funding/player-a/state.ts
```diff
@@ -49,33 +44,50 @@ export type PlayerAState =
...
   const { channelId, ledgerId } = params;
-  return { type: WAIT_FOR_LEDGER_UPDATE_1, player: PlayerIndex.A, channelId, ledgerId };
+  return { type: WAIT_FOR_LEDGER_UPDATE_1, channelId, ledgerId };
+}
+
+// -------
+// Helpers
+// -------
+
+export function isPlayerAState(state: IndirectFundingState): state is PlayerAState {
```
lalexgap: One (minor) downside is that we have another type guard that needs to be updated whenever we add/remove a state.

tomclose: That's true. I guess that you could catch that by always using these guards with exhaustive checks:
```ts
// instead of this:
if (isPlayerAState(state)) {
    // do player A stuff
} else {
    // do player B stuff
}

// do this:
if (isPlayerAState(state)) {
    // do player A stuff
} else if (isPlayerBState(state)) {
    // do player B stuff
} else {
    unreachable(state)
}
```
In fact, one test like that would ensure that the tests would catch it if you ever missed a state.

Overall it seems a bit borderline as to whether this is an improvement or not. I guess it's useful to be aware of it as an alternative.

PR #321, opened 2019-04-08
Wallet architecture proposal

tomclose:
This PR proposes a solution to the questions that came up at the end of last week around how processes in the wallet should be composed.

### DIFF
--- a/packages/wallet/notes/structuring-the-wallet.md
+++ b/packages/wallet/notes/structuring-the-wallet.md
@@ -1,0 +1,151 @@
+# Wallet Architecture
+
+The wallet needs to be able to run many different protocols, potentially simultaneously.
+For example, the wallet might need to be funding channel C1 with opponent A at the same time
+as responding to a challenge on channel C2.
+As we introduce state channel networks, some protocols will need involve interactions between
+multiple channels. This means tha we must move beyond our approach to-date, where the
+protocol was recorded as part of the channel state.
+
+It is clear that the different protocols will have many areas of overlap.
+Examples here include (a) submitting a transaction to the blockchain, which is part of
+many different protocols and (b) directly funding a channel, which occurs independently
+but can also occur as part of indirect funding, if we need to directly fund the ledger channel.
+We need to find an architecture that allows us to reuse the shared logic in these cases,
+while remaining simple and easy to understand.
+
+This document gives a proposal for such an architecture.
+
+## Protocols and Protocol composition
+
+A **protocol** defines the series of interactions that must take place between participants
+and the blockchain in order to accomplish a given outcome.
+Examples of protocols include "funding a channel", "launching a challenge" and "withdrawing
+funds".
+The purpose of a wallet is to provide an _implementation_ for the protocols defined by the
+framework.
+The wallet provides this implementation by defining a state machine, where the states track
+the progress made in executing the protocol.
+
+As mentioned before, there is a lot of structure and overlap between the different protocols.
+In the wallet, we represent this structure by embedding state machines inside each other.
+For example, the state machine for submitting a transaction to the blockchain is as follows:
+
+```mermaid
+graph LR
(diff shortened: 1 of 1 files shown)


lalexgap:
This approach sounds good to me! I think we should spend some time breaking this down into different chunks of work so we're all on the same page.

tomclose:
@andrewgordstewart not sure what you mean here. The actions for a protocol will depend on the state machine that implements the protocol.

andrewgordstewart:
@tomclose I mean is there a specific shape that the process' action needs to look like? It clearly needs to extend something like `{ processId: any }`. Are there any other requirements?

tomclose:
@andrewgordstewart are you happy to approve this, so that we can merge it in?

andrewgordstewart:
I'm happy to merge it, but how actions behave is still undetermined.

## Example 
1. Player A requests indirect funding
2. Wallet A starts an `IndirectFunding` process with id `0`
3. Player B requests indirect funding
4. Maybe Wallet B already has some process with id `0` going on, so it starts an `IndirectFunding` process with id `1`

Now, the process actions are out of sync.

This example raises some questions:
- Should we allow this to happen?
  * If yes, how do we handle this situation? 
  * If no, how do we prevent it?
- Should the `FundingRequested` action have a `processId` on it? (ie. does every "process" action need a `processId`?)
- What does the wallet do when it receives an action with a `processId` on it, but it doesn't have a process running with that id?

tomclose:
I think there are a couple of options here.

One way this could work is that there's always an 'instigator'. E.g. in the funding case, you can break the symmetry based on the order of participants in the channel you're funding. The first participant proposes indirect funding and picks a `processId`. This could even be prefaced with the hash of the participants to avoid collisions. (You'll remember that the original design had something like this - where player A's wallet proposed indirect funding and player B's wallet accepted.)

Another way would be to make the `processId` somehow deterministic. E.g. if you're funding channel id 'abc123' then it could be as simple as making the processid = `fund.abc123` or something.

I think there will probably end up being a set of actions (e.g. requests from the app and proposals from other participants) which won't have a `processId` but will cause the wallet to create a process. If the wallet receives an action that belongs to a process that the wallet doesn't have, it should probably(?) ignore it. To me this seems similar to receiving an action that you're not in the right state for. In the long term we might want to respond with an 'unexpected action' error or similar.

andrewgordstewart:
> I think there will probably end up being a set of actions (e.g. requests from the app and proposals from other participants) which won't have a `processId` but will cause the wallet to create a process.

So, I'm wondering if there should be "protocol" actions (actions that have a `protocol` property) and "process" actions (actions that have a `processId` property). Process actions are necessarily protocol actions, but some protocol actions are used to trigger a new process. (Ie. protocol actions are partitioned to "active-process" actions and "new-process" actions)

> If the wallet receives an action that belongs to a process that the wallet doesn't have, it should probably(?) ignore it. To me this seems similar to receiving an action that you're not in the right state for. In the long term we might want to respond with an 'unexpected action' error or similar.

However, it may be the case that the instigator (A) chooses the process ID, sends an action to B that includes that `processId` -- and this is how they're kept in sync -- and then B's Wallet has to route an action that has a `processId`, but with no process running. In this case, there would be no nice way to categorize  "active-process " actions and "new-process" actions

tomclose:
It seems that the set of actions will be partitioned with 'existing-process actions' and 'new-process actions' forming two of the partitions. We might want to call new-process actions 'protocol actions' and enforce that they have a `protocol` property but there are also other ways of achieving the same ends. In some cases (e.g. wallet A proposing funding to wallet B), the 'new-process' action will have a `processId` - but that's ok, as the wallet will understand that the instruction is to create a new process. (That action will need to contain all the details of the process that is to be created). If an existing-process action is received and there is no corresponding process, then that's the case where the action will be ignored.

Thinking about it more, I'm beginning to thing that deterministic process ids might be the way to go. (Though it requires more thought, as it isn't clear that we'll always have a unique piece of information.)

PR #323, opened 2019-04-08
Player A Reducer tests

lalexgap:


packages/wallet/src/redux/channel-state/reducer.ts
```diff
@@ -29,6 +29,7 @@ import { StateWithSideEffects } from '../utils';
 import { ethers } from 'ethers';
 import { channelID } from 'fmg-core/lib/channel';
 import { WalletAction, COMMITMENT_RECEIVED } from '../actions';
+import { Commitment } from 'fmg-core/lib/commitment';
```
andrewgordstewart: shouldn't this be imported from `fmg-core` itself?

packages/wallet/src/redux/channel-state/reducer.ts
```diff
@@ -353,3 +354,9 @@ const receivedValidOpponentConclusionRequest = (
     penultimateCommitment: state.lastCommitment,
   });
 };
+
+const channelIsInitialized = (commitment: Commitment, state: states.ChannelState): boolean => {
+  const channelId = channelID(commitment.channel);
+  const initializedChannelAddresses = new Set(Object.keys(state.initializedChannels));
```
andrewgordstewart:
```suggestion
  return channelId in state. initializedChannels
```

packages/wallet/src/redux/channel-state/reducer.ts
```diff
@@ -353,3 +354,9 @@ const receivedValidOpponentConclusionRequest = (
...
+  return initializedChannelAddresses.has(channelId);
```
andrewgordstewart:
```suggestion

```

packages/wallet/src/redux/indirect-funding/player-a/__tests__/reducer.test.ts
```diff
@@ -0,0 +1,238 @@
...
+  walletState.indirectFunding = states.waitForPreFundSetup1({ channelId, ledgerId });
+  // Add the ledger channel to state
+  const ledgerChannelState = channelStates.waitForPreFundSetup({
+    ...ledgerChannelDefaults,
+    channelId: ledgerId,
+  });
+  walletState.channelState.initializedChannels[ledgerId] = ledgerChannelState;
+
+  describe(whenActionArrives(actions.COMMITMENT_RECEIVED), () => {
+    const validateMock = jest.fn().mockReturnValue(true);
```
andrewgordstewart: Can you move this mock to the top of the test suite?

packages/wallet/src/redux/indirect-funding/player-a/__tests__/reducer.test.ts
```diff
@@ -0,0 +1,238 @@
...
+    itTransitionsChannelToStateType(
+      updatedState,
+      newLedgerId,
+      channelStates.WAIT_FOR_PRE_FUND_SETUP,
+    );
+  });
+});
+
+describe(startingIn(states.WAIT_FOR_PRE_FUND_SETUP_1), () => {
+  const { channelId, ledgerId } = defaults;
```
andrewgordstewart: 💡 Lines 112-20 seems common among a lot of test cases, and might want to be extracted into something that constructs a starting state. This way, when we change the indirect funding reducer's signature, we only have to update the "starting state" construction in one place.

packages/wallet/src/redux/indirect-funding/player-a/__tests__/reducer.test.ts
```diff
@@ -0,0 +1,238 @@
...
+    const validateMock = jest.fn().mockReturnValue(true);
+    Object.defineProperty(SigningUtil, 'validCommitmentSignature', { value: validateMock });
+
+    const action = actions.commitmentReceived(
+      ledgerId,
+      WalletProcedure.IndirectFunding,
+      testScenarios.ledgerCommitments.ledgerUpdate1,
+      '0x0',
+    );
+    playerAReducer(walletState, action);
```
andrewgordstewart: Is the purpose of this describe block is to show that there's no run-time error?

packages/wallet/src/redux/channel-state/reducer.ts
lalexgap: Yes, for some reason vscode auto import seems to want to pull it off lib.

packages/wallet/src/redux/indirect-funding/player-a/__tests__/reducer.test.ts
lalexgap: Yup, right now it only serves to make sure things don't blow up. Once we decide on what a "finished" state looks like this can be modified to do the proper check.

packages/wallet/src/redux/channel-state/reducer.ts
lalexgap: Also it looks like an issue we have in a lot of places. We should probably go through and clean it up at some point.

packages/wallet/src/redux/indirect-funding/player-a/__tests__/reducer.test.ts
lalexgap: I think that's a good idea. I'm going to defer that until we refactor the reducer tests based on the Protocol Reducer interface.

packages/wallet/src/redux/indirect-funding/player-a/__tests__/reducer.test.ts
lalexgap: Done

packages/wallet/src/redux/indirect-funding/player-a/__tests__/reducer.test.ts
andrewgordstewart: Yeah, I would do it at the same time as that refactor.

PR #324, opened 2019-04-08
Switch Message actions to use processId

lalexgap:
Switch `COMMITMENT_RECEIVED`/`MESSAGE_RECEIVED` to use `processId`.

This allows the `COMMITMENT_RECEIVED` action to be passed into the `channelStateReducer` and have it correctly deduce the channelId

### DIFF
--- a/packages/wallet/src/redux/actions.ts
+++ b/packages/wallet/src/redux/actions.ts
@@ -46,6 +46,7 @@ export const metamaskLoadError = () => ({
 export type MetamaskLoadError = ReturnType<typeof metamaskLoadError>;
 
 // Common Transaction Actions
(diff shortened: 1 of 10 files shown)


packages/wallet/src/redux/actions.ts
```diff
@@ -46,6 +46,7 @@ export const metamaskLoadError = () => ({
 export type MetamaskLoadError = ReturnType<typeof metamaskLoadError>;
 
 // Common Transaction Actions
+// TODO: These should be switched from channelId to processId
```
lalexgap: Right now `directFunding` is expecting a `channelId` on the the transaction actions so I'm leaving this in place for now.

PR #328, opened 2019-04-09
Adjudicator Watcher now Dispatches actions with a process Id

### LINKED ISSUE #325: Adjudicator-Watcher needs to dispatch actions with proccessId
Currently the `adjudicator-watcher` is registered against a single channelID and the actions that are dispatched have a channelId.

The `adjudicator-watcher` will need to dispatch actions with a `processId` so they can be dispatched to the correct process and protocol.

We need a way of registering with the `adjudicator-watcher` the channels that the process (or protocol?) is interested in.

We could have something like this: `listenForAdjudicatorEvents(channelIds:string[], processId:string)` that adds an entry to a registry `{[channelId:string]: proccessIds:string[] }` in `ajudicator-watcher`. 

Whenever an adjudicator event is detected by the `adjudicator-watcher` it looks up any processIds interested in that channel and dispatches an action like `fundingReceivedEvent(processId:string, channelId:string, amount:string, amountForDestination:string)`

lalexgap:
The adjudicator-watcher now listens for a `REGISTER_FOR_ADJUDICATOR_EVENTS` or `UNREGISTER_FOR_ADJUDICATOR_EVENTS` that a process can use to register for events for a list of channels.

The `adjudicator-watcher` will dispatch event actions with a processId if a process has registered for the given channelId.

I've moved the actions around a little bit but I think we need to do a more thorough refactoring after we've completed some of the other work (such as moving DirectFunding into a procedure)

resolves #325

### DIFF
--- a/packages/wallet/src/contract-tests/adjudicator-events.test.ts
+++ b/packages/wallet/src/contract-tests/adjudicator-events.test.ts
@@ -9,6 +9,7 @@ import {
   refuteChallenge,
   respondWithMove,
   getChannelId,
+  defaultDepositAmount,
 } from './test-utils';
 
 jest.setTimeout(60000);
@@ -21,6 +22,7 @@ describe('adjudicator listener', () => {
   const participantA = ethers.Wallet.createRandom();
   const participantB = ethers.Wallet.createRandom();
   let nonce = 5;
+  const processId = 'A process identifier';
   function getNextNonce() {
     return ++nonce;
   }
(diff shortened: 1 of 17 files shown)


packages/wallet/src/contract-tests/adjudicator-events.test.ts
```diff
@@ -31,91 +33,156 @@ describe('adjudicator listener', () => {
...
     );
-    await sagaTester.waitFor(actions.channel.RESPOND_WITH_MOVE_EVENT);
-    const action: actions.channel.RespondWithMoveEvent = sagaTester.getLatestCalledAction();
-    expect(action.type === actions.channel.RESPOND_WITH_MOVE_EVENT);
-    expect(action.responseCommitment).toEqual(responseState);
+
+    await sagaTester.waitFor(actions.RESPOND_WITH_MOVE_EVENT);
+
+    const action: actions.RespondWithMoveEvent = sagaTester.getLatestCalledAction();
+    expect(action).toEqual(actions.respondWithMoveEvent(processId, channelId, responseCommitment));
```
andrewgordstewart: We should probably have tests for re-registering and cancelling, right?

PR #331, opened 2019-04-10
DirectFunding protocol

andrewgordstewart:


packages/wallet/src/redux/direct-funding/depositing/reducer.ts
```diff
@@ -1,29 +1,49 @@
...
+      return applyUpdate(state, waitForDepositConfirmationReducer(protocolState, action));
     case states.DEPOSIT_TRANSACTION_FAILED:
-      return depositTransactionFailedReducer(state, action);
+      return applyUpdate(state, depositTransactionFailedReducer(protocolState, action));
   }
-  return unreachable(state);
+  return unreachable(protocolState);
 };
 
+function applyUpdate(
```
lalexgap: I think it'd be nice to change the subReducers to take a `ProtocolStateWithSharedData` so we don't need this helper function and makes the reducer easier to understand.

Longer term that will allow us to `StateWithSideEffects` and hopefully simplify the `SideEffects`

PR #334, opened 2019-04-11
Change Player A&B reducers to implement ProtocolReducer interface.

lalexgap:
Updates Player A Reducer, Player B Reducer,Indirect Funding reducer, and the shared reducer utils to implement the `ProtocolReducer` interface.


packages/wallet/src/redux/indirect-funding/player-a/reducer.ts
```diff
@@ -32,180 +30,191 @@ import {
...
-    return state;
-  }
-
-  switch (state.indirectFunding.type) {
+): ProtocolStateWithSharedData<states.PlayerAState> {
+  switch (state.protocolState.type) {
     case states.WAIT_FOR_APPROVAL:
-      return waitForApprovalReducer(state, action);
+      return waitForApprovalReducer(
+        state as ProtocolStateWithSharedData<states.WaitForApproval>,
```
andrewgordstewart: This should not require a type assertion. Either the input type to `waitForApprovalReducer` should be relaxed, or we should figure out how to properly do the discriminated union.

packages/wallet/src/redux/selectors.ts
```diff
@@ -1,24 +1,18 @@
...
+import { OpenedState, OPENING, ChannelStatus, ChannelState } from './channel-state/state';
 import * as walletStates from './state';
 import * as indirectFundingStates from './indirect-funding/state';
 import { DirectFundingState } from './direct-funding/state';
 
-export const getOpenedChannelState = (
-  state: walletStates.Initialized,
-  channelId: string,
-): OpenedState => {
+export const getOpenedChannelState = (state: ChannelState, channelId: string): OpenedState => {
```
andrewgordstewart: I think the `state` variable should be as close to a wallet state as possible. It can, and probably should be of type `{sharedData: SharedData}`

packages/wallet/src/redux/indirect-funding/player-a/reducer.ts
tomclose: Something that would make this easier would be to change the signature of these reducers to:
```ts
function waitForApprovalReducer(
    protocolState: WaitForApproval,
    sharedState: SharedState,
    action: IndirectFundingAction
): ReturnVal {
  // ...
}

type ReturnVal = ProtocolStateWithSharedData<InsufficientFundingState>;
```

packages/wallet/src/redux/selectors.ts
lalexgap: I'm a bit hesitant about this since it infers our `walletState` has the same shape as our `ProtocolWithSharedData`.

I prefer having the selectors more strongly typed against a specific type of state.

andrewgordstewart: Do you mean then that a selector that wants a specific channel state, given a wallet state, should call
`getOpenedChannelState(getChannelState(walletState), channelId)`? Or that they should call `getOpenedChannelState(walletState.channelState, channelId)`?

The second option goes against the philosophy that callers of a selector need to know something about the shape of the wallet state. This is ok if it's a "private" selector that's called in the context where you directly have a `channelState` object, but it shouldn't be used in a UI component, for instance.

PR #340, opened 2019-04-11
Transaction submission

tomclose:
This PR is the first step towards splitting transaction submission out into its own protocol. The work is still in progress, but I wanted to get this part reviewed and merged, so that the PR doesn't become too large.

Still todo in a future PR:

- [ ] Queue the transaction in the outbox
- [ ] Add tests
- [ ] Add components/containers
- [ ] Add storybook


packages/wallet/src/redux/protocols/transaction-submission/reducer.ts
```diff
@@ -0,0 +1,99 @@
...
+
+function transactionSentToMetamask(state: TSState, storage: Storage): ReturnVal {
+  if (state.type !== 'Start') {
+    return { state, storage };
+  }
+  return { state: waitForSubmission(state), storage };
+}
+
+function transactionSubmissionFailed(state: TSState, storage: Storage): ReturnVal {
+  if (state.type !== 'WaitForSubmission') {
```
andrewgordstewart: These should be constants, not magic strings.

PR #343, opened 2019-04-11
Move indirect funding to protocols folder

lalexgap:
Moves the indirect-funding folder to the protocols folder to follow the same format as other protocols.


packages/wallet/src/redux/selectors.ts
```diff
@@ -20,16 +19,6 @@ export const getChannelState = (state: SharedData, channelId: string): ChannelSt
   return channelStatus;
 };
 
-// TODO: Ideally we should be able to pass in a expected state type and have the selector either return that type of state or throw an error
-export function getIndirectFundingState(
```
andrewgordstewart: Why was this removed?

lalexgap: It's been obsoleted now that the reducers are handling `protocolState` directly. It should have been removed earlier but I just caught it now.

PR #345, opened 2019-04-12
Indirect Funding now starts Direct Funding protocol and stores it in state

lalexgap:
Indirect Funding now has DirectFunding protocol state when waiting for direct funding.

The indirect funding reducer also handles initializing the direct funding protocol and routes funding actions to it.


packages/wallet/src/redux/__tests__/test-scenarios.ts
```diff
@@ -228,3 +233,24 @@ export const ledgerCommitments = {
...
+  const requiredDeposit = twoThree[ourIndex];
+
+  const action = actions.internal.directFundingRequested(
+    fundingRequestChannelId,
+    safeToDepositLevel,
+    total,
+    requiredDeposit,
+    ourIndex,
+  );
+  return directFundingStates.initialDirectFundingState(action, EMPTY_SHARED_DATA).protocolState;
```
andrewgordstewart: It seems strange to me that you would use wallet code to construct a test scenario. Why not construct it explicitly?

packages/wallet/src/redux/protocols/direct-funding/state.ts
```diff
@@ -105,7 +108,10 @@ export type DirectFundingState =
   | WaitForFundingConfirmation
   | ChannelFunded;
 
-export function initialDirectFundingState(action: DirectFundingRequested) {
+export function initialDirectFundingState(
```
andrewgordstewart: I wonder if we should make a convention around naming. In the `TransactionSubmission` protocol, there is an `initialize` function, which seems reasonable.

packages/wallet/src/redux/protocols/indirect-funding/player-a/reducer.ts
```diff
@@ -117,15 +120,22 @@ const waitForDirectFunding = (
...
+    );
+
+    let newSharedData = updatedStateAndSharedData.sharedData;
+    let newProtocolState: states.PlayerAState = updatedStateAndSharedData.protocolState;
+
+    if (directFundingIsComplete(newProtocolState)) {
+      newSharedData = confirmFundingForChannel(sharedData, protocolState.ledgerId);
       newSharedData = createAndSendPostFundCommitment(newSharedData, protocolState.ledgerId);
-      const newProtocolState = states.waitForPostFundSetup1(protocolState);
+      newProtocolState = states.waitForPostFundSetup1(updatedStateAndSharedData.protocolState);
```
andrewgordstewart:
```suggestion
      newProtocolState = states.waitForPostFundSetup1(newProtocolState);
```

packages/wallet/src/redux/protocols/indirect-funding/player-a/state.ts
```diff
@@ -16,6 +17,10 @@ interface LedgerChannelExists extends BasePlayerAState {
   ledgerId: string;
 }
 
+interface DirectFundingExists extends LedgerChannelExists {
```
andrewgordstewart: I don't think this interface is needed, since it will only ever be used once. Therefore, `WaitForDirectFunding` can extend `LedgerChannelExists`, and simply add a `directFundingState` property itself.

packages/wallet/src/redux/protocols/indirect-funding/player-b/__tests__/reducer.test.ts
```diff
@@ -140,30 +142,30 @@ describe(startingIn(states.WAIT_FOR_PRE_FUND_SETUP_0), () => {
...
-    itSendsNoMessage(updatedState.sharedData);
+    expectThisCommitmentSent(updatedState.sharedData, ledgerCommitments.preFundCommitment1);
     itSendsNoTransaction(updatedState.sharedData);
   });
 });
 
 describe(startingIn(states.WAIT_FOR_DIRECT_FUNDING), () => {
-  describe.skip(whenActionArrives(actions.FUNDING_RECEIVED_EVENT), () => {
-    // Need to hook up the direct funding store first, which isn't yet in this branch
+  describe(whenActionArrives(actions.FUNDING_RECEIVED_EVENT), () => {
```
andrewgordstewart: :tada:

packages/wallet/src/redux/protocols/indirect-funding/player-a/reducer.ts
```diff
@@ -181,10 +189,41 @@ const waitForApprovalReducer = (
   }
 };
 
+const updateStateWithDirectFundingAction = (
```
andrewgordstewart: This function is identical to the function with the same name defined below.

packages/wallet/src/redux/protocols/indirect-funding/player-a/reducer.ts
```diff
@@ -181,10 +189,41 @@ const waitForApprovalReducer = (
...
+    sharedData: updatedSharedData,
+  } = requestDirectFunding(sharedData, protocolState.ledgerId);
+  const newProtocolState = states.waitForDirectFunding({
+    ...protocolState,
+    directFundingState: directFundingProtocolState,
+  });
+  return { protocolState: newProtocolState, sharedData: updatedSharedData };
+};
+
 const directFundingIsComplete = (protocolState: states.WaitForDirectFunding): boolean => {
```
andrewgordstewart: This should be common between player a and b.

packages/wallet/src/redux/protocols/indirect-funding/reducer-helpers.ts
```diff
@@ -191,20 +213,22 @@ export const updateChannelState = (
...
   return newSharedData;
 };
 
 export const updateDirectFundingStatus = (
   directFundingState: DirectFundingState,
+  sharedData: SharedData,
   action: FundingAction,
-): DirectFundingState => {
+): ProtocolStateWithSharedData<DirectFundingState> => {
   // TODO: Update the direct funding state
```
andrewgordstewart:
```suggestion

```

packages/wallet/src/redux/protocols/indirect-funding/player-a/reducer.ts
lalexgap: I'm not sure I follow, which function is identical to this?

packages/wallet/src/redux/__tests__/test-scenarios.ts
lalexgap: Mostly convenience over having to manually recreate and maintain a direct funding state.

PR #346, opened 2019-04-12
Add section on actions

andrewgordstewart:
Proposes one way that the wallet reducer can handle actions for new and ongoing actions in a predictable manner.


packages/wallet/notes/structuring-the-wallet.md
```diff
@@ -148,4 +148,60 @@ There needs to be some top level handler whose responsibility it is to trigger p
...
+The one time when it makes sense for `a` to store a protocol is if
+
+- we're not in a system where the process id is a function of the action's data
+- there is no process running with id `a.processId`
+
+In this case, when my wallet sends `a` to your wallet, your wallet understands this to mean
+
+> _start a process with id `a.processId`, running protocol `a.protocol`_
+
+By making the process id deterministic, we rule out the need for this, making actions easy to reason about:
```
tomclose: I don't understand why the determinism affects this. It seems that we can have this behaviour regardless of how the process id is chosen.

tomclose: (Here I'm allowing for new-process actions to have a property specifying the process id for the new process in some cases e.g. via a `newProcessId` or even just `processId` property.)

andrewgordstewart: Determinism isn't required to have this behaviour. This behaviour is a consequence of determinism. I'm arguing that we want this behaviour -- I probably should have lead with that -- and that deterministic ids is one way to achieve this.

To be clear: We want actions that are very clearly meant to _either_
A. create a new process
B. be routed to an existing process
and having `processId` precisely on actions covered by `B` is one way to achieve this.

andrewgordstewart: (And, having a `protocol` property on actions covered by `B` is redundant/misleading)

tomclose: I agree that we want that behaviour. I'm anticipating that the `type` property of the action will completely determine whether it is a new-process or existing-process action. Does that fit with what you're thinking?

If it's true that `type` determines its class, it's then just an engineering decision on how we want to select members of that type. Presence of a `processId` is probably a good choice. If we do that then we'd probably need to allow for a `proposedProcessId` on some new-process actions but that's fine.

andrewgordstewart:
Superceded by https://github.com/magmo/apps/pull/354

PR #347, opened 2019-04-12
Transaction submission

tomclose:


packages/wallet/src/redux/protocols/transaction-submission/reducer.ts
```diff
@@ -1,30 +1,34 @@
...
-import { TransactionRequest } from 'ethers/providers';
 import { unreachable } from '../../../utils/reducer-utils';
 import { SharedData } from '..';
+import { TransactionRequest } from 'ethers/providers';
+import { queueTransaction } from '../../state';
 
 type Storage = SharedData;
 
-interface ReturnVal {
+export interface ReturnVal {
```
lalexgap: I'm not a huge fan of exporting something called`ReturnVal`. It seems a bit generic.

PR #351, opened 2019-04-12
Generalizing wallet checklists

kerzhner:
A wallet checklist is an ordered set of steps with completed steps, in-progress step, and uncompleted steps. An example for indirect funding is: 
Screen Shot 2019-04-12 at 3 19 40 PM

This PR is an attempt to generalize wallet checklists into a reusable component. Clients of the component need to define a `Step` `enum` and a set of messages for each step. Right now, there is no static check for ensuring that each step in the enum has a corresponding set of messages. This can lead to a runtime error.

For example, if we add a new `Step` constant to the end of the list and forget to add a set of messages to the `messagesForStepList` list for the new step, there can be a runtime error.


packages/wallet/src/components/indirect-funding/funding-step.tsx
```diff
@@ -63,80 +42,45 @@ export const fundingStepByState = (state: indirectFunding.IndirectFundingState):
...
-        return 'Not ready to allocate from ledger channel';
-      } else if (currentStep === iconStep) {
-        return 'Waiting for opponent ledger channel update';
-      } else {
-        return 'Received opponent ledger channel update';
-      }
-  }
-  return '';
-};
+const messagesForStepList: MessagesForStep[] = [
```
andrewgordstewart: Maybe a better option here is to have an object that has a key for every step in the `Step` enum on line 12. 

~You can get some type safety by defining a `MessagesForStep` interface that looks like `{ [step: Step]: MessageForStep }`~ Sadly, `Step` is not a valid type for an index signature.

If you really want type safety, you could have a `messagesForStepList` _function_, that has an exhaustive `switch` statement on the `Step` enum.

packages/wallet/src/components/indirect-funding/funding-step.tsx
```diff
@@ -63,80 +42,45 @@ export const fundingStepByState = (state: indirectFunding.IndirectFundingState):
...
-            {message(Step.WAIT_FOR_POST_FUND_SETUP, currentStep)}
-          </li>
-          <li>
-            {icon(Step.WAIT_FOR_LEDGER_UPDATE, currentStep)}
-            {message(Step.WAIT_FOR_LEDGER_UPDATE, currentStep)}
-          </li>
-        </ul>
-        <div className="pb-2">{children}</div>
-      </StatusBarLayout>
+      <Checklist
```
andrewgordstewart: 🌋

packages/wallet/src/components/indirect-funding/funding-step.tsx
kerzhner: I did try out the `messagesForStepList` function approach earlier and didn't think the additional complexity was worth the additional safety.

PR #353, opened 2019-04-13
Sketch out how to route actions

andrewgordstewart:
This PR does not actually route the actions, but sets up a framework for how they should be routed:
A. if an action is meant to create a new process (right now, that means it has a `protocol` property) then it is routed to the corresponding protocol's "initializer"
B. if an action is meant for a running process, that process' state `s` is pulled from the store, and the protocol reducer is applied to `(s, sharedData, action)`

## Future work
- [ ] Clean up existing actions to fit this paradigm, as needed.
  * There might need to be some convention on how a 
- [ ] Define initializers for the various protocols. (There is already an initializer defined in the `TransactionSubmission` protocol.)
  * This probably requires some additional "new-process" actions.
- [ ] Actually call out to the various protocol reducer


packages/wallet/src/redux/protocols/transaction-submission/actions.ts
```diff
@@ -1,52 +1,60 @@
...
+  protocol: WalletProtocol.TransactionSubmission;
   channelId: string;
   processId: string;
   transactionHash: string;
 }
 
-export interface TransactionConfirmed {
+export interface TransactionConfirmed extends BaseProcessAction {
   type: typeof TRANSACTION_CONFIRMED;
+  protocol: WalletProtocol.TransactionSubmission;
```
lalexgap: Can we use a shared common interface for the shared properties? Or have we decided we want to redeclare the properties on each interface?

packages/wallet/src/redux/reducer.ts
```diff
@@ -36,6 +37,13 @@ export function initializedReducer(
   state: states.Initialized,
   action: actions.WalletAction,
 ): states.WalletState {
+  if (createsNewProcess(action)) {
```
lalexgap: It would be good to start stubbing out tests for these. Even if there's not much to test yet.

packages/wallet/src/redux/protocols/actions.ts
```diff
@@ -0,0 +1,28 @@
+import { WalletProtocol } from '../types';
+import { WalletAction } from '../actions';
+import { ProcessAction as IndirectFundingProcessAction } from './indirect-funding/actions';
+
+export interface NewProcessAction {
+  protocol: WalletProtocol;
+}
+
+export function createsNewProcess(action: WalletAction) {
```
lalexgap:
```suggestion
export function createsNewProcess(action: WalletAction) : boolean {
```

packages/wallet/src/redux/reducer.ts
andrewgordstewart: I actually tried to! I still don't feel like I know how to mock things :/

I will try again.

packages/wallet/src/redux/protocols/transaction-submission/actions.ts
andrewgordstewart: This is cherry-picked from Tom's branch -- the protocol should actually be removed from these actions, since they have a process id. (The protocol lives on the process state)

packages/wallet/src/redux/reducer.ts
andrewgordstewart: I'll leave the tests out of this PR -- later, I will actually call out to the protocol reducers, and will stub those out.

packages/wallet/src/redux/protocols/transaction-submission/actions.ts
```diff
@@ -1,52 +1,60 @@
...
+export const TRANSACTION_SUBMISSION_FAILED = 'WALLET.TRANSACTION_SUBMISSION_FAILED';
+export const TRANSACTION_SUBMITTED = 'WALLET.TRANSACTION_SUBMITTED';
+export const TRANSACTION_CONFIRMED = 'WALLET.TRANSACTION_CONFIRMED';
+export const RETRY_TRANSACTION = 'WALLET.RETRY_TRANSACTION';
+export const TRANSACTION_FINALIZED = 'WALLET.TRANSACTION_FINALIZED';
 
-export interface TransactionSentToMetamask {
+export interface TransactionSentToMetamask extends BaseProcessAction {
   type: typeof TRANSACTION_SENT_TO_METAMASK;
+  protocol: WalletProtocol.TransactionSubmission;
```
tomclose: I'm confused as to why these actions have both a `protocol` and a `processId`. In any case, it seems that the `protocol` here can be deduced from the action type. Maybe I don't understand how this `protocol` property will be used.

packages/wallet/src/redux/protocols/transaction-submission/actions.ts
```diff
@@ -1,52 +1,60 @@
...
   channelId: string;
   processId: string;
 }
-export interface TransactionSubmissionFailed {
+export interface TransactionSubmissionFailed extends BaseProcessAction {
```
tomclose: On reading `BaseProcessAction` I see that is has a single property: `processId`. It makes sense to me that `TransactionSubmissionFailed` would _implement_ `BaseProcessAction` (i.e. conform to its interface) but it doesn't need to extend `BaseProcessAction` to do this - in fact it already has this property below. I don't think the tradeoff of having to look up another interface in another file makes sense in this case.

tomclose:
I wrote up the problem I'm talking about here: https://github.com/magmo/apps/pull/354.

packages/wallet/src/redux/protocols/transaction-submission/actions.ts
andrewgordstewart: I believe this is from a commit that I cherry-picked from your branch, and I opted not to remove the `protocol` property quite yet. In fact, the proposal suggests _never_ having a `protocol` and `processId` property, which these interfaces have.

packages/wallet/src/redux/protocols/transaction-submission/actions.ts
andrewgordstewart: It doesn't make sense for an interface to _implement_ another interface. An interface is a contract -- classes can implement a contract, and interfaces can extend on other contracts.

I don't really understand the resistance towards extending interfaces as a means of documentation and clarity. "A process action needs to have a `processId`" is a valid thing to be explicit about. Of course, you can be explicit about that in the definition of a `isProcessAction` (naming?) typeguard. However, that's a different issue than being explicit about the fact that `TransactionSubmissionFailed` _needs to be a process action_. That's the purpose of extending a common interface.

Without this, the connection between the `TransactionSubmissionFailed` interface and the definition of what it means to be a process action is lost: there is no connection made in the codebase, only in the head of a developer that manages to keep in their head the definition of `isProcessAction`, at all times.

Keeping the redundant `processId` property was in an effort to make it easier to see what is on an object of type `TransactionSubmissionFailed`, like you want to do by not extending something like the `BaseProcessAction`.

IMO, this is not great: if you want to rename `processId` to `protocolLocator`, for instance, it's so much easier to do so if every process action simply extends a `BaseProcessAction` interface. Then you rename it in two places: once in the base interface, and once in the type guard. The current approach lets you get the effect you want with just those two renames, with the negative side effect that you end up with a `processId` on all of these interfaces that you wanted to be explicit about. To me, that still seems better than just having `processId` on everything, with no common definition of what a "process action is" -- then, you have to search and replace across the codebase and hope you don't make a mistake.

packages/wallet/src/redux/protocols/transaction-submission/actions.ts
andrewgordstewart: Not being sure where the `protocol` came from, I've removed it.

PR #372, opened 2019-04-16
Transaction submission views

tomclose:
These views are very basic - more of a placeholder until we figure out what we want the overall user experience to be.

transaction-submission

### DIFF
--- a/packages/wallet/src/redux/protocols/transaction-submission/components/wait-for-confirmation.tsx
+++ b/packages/wallet/src/redux/protocols/transaction-submission/components/wait-for-confirmation.tsx
@@ -1,0 +1,40 @@
+import React, { Fragment } from 'react';
+
+interface Props {
+  name: string;
+  transactionId: string;
+  networkId: number;
+}
+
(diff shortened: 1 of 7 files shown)


packages/wallet/src/redux/protocols/transaction-submission/components/wait-for-confirmation.tsx
```diff
@@ -0,0 +1,40 @@
...
+interface Props {
+  name: string;
+  transactionId: string;
+  networkId: number;
+}
+
+export default class WaitForXConfirmation extends React.PureComponent<Props> {
+  buildEtherscanLink() {
+    switch (this.props.networkId) {
+      case 1:
```
kerzhner: @tomclose This might be cleaner with an Enum

tomclose: Yeah - that's worth looking at but probably as future work. In this case, I just did a copy-and-paste from the code that was already in the `src/components` folder ...

PR #377, opened 2019-04-16
Withdrawal Protocol

lalexgap:
Adds the withdrawing protocol with the following:
- Readme with state machine
- states/actions
- scenarios
- Reducer and reducer tests
- views (pretty rough)
- storybook


packages/wallet/src/redux/protocols/withdrawing/states.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export interface Success {
+  type: typeof SUCCESS;
+}
+
+export interface Failure {
+  type: typeof FAILURE;
+  reason: string;
+}
```
tomclose: Duplicate definition from line 38.

packages/wallet/src/redux/protocols/withdrawing/states.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export function success(): Success {
+  return { type: SUCCESS };
+}
+
+export function failure(reason: string): Failure {
+  return { type: FAILURE, reason };
+}
+
+export const FAILURE_REASONS = {
```
tomclose: This is a nice improvement over the `TransactionSubmission` protocol. Could it be an enum though? That way you could use type safety to limit the allowed failure reasons in the `Failure` state.

packages/wallet/src/redux/protocols/withdrawing/readme.md
```diff
@@ -0,0 +1,34 @@
...
+
+```mermaid
+graph LR
+  St((start)) --> WFAp(WaitForApproval)
+  WFAp --> |Approved| WFT(WaitForTransaction)
+  WFT --> |TransactionSubmitted| WFAk(WaitForAcknowledgement)
+  WFT --> |TransactionFailed| F((failure))
+  WFAk --> |SuccessAcknowledged| S((success))
+  WFAp --> |Rejected| F((failure))
+```
```
tomclose: I think you need at least one more state here to handle the situation where the user tries to withdraw before the channel is closed (e.g. either a conclusion proof exists or the channel has been finalized on-chain). It probably makes sense to dump the user straight into that state instead of `WaitForApproval`. It would probably be `AcknowledgeNotClosed` or something.

packages/wallet/src/redux/protocols/withdrawing/readme.md
```diff
@@ -0,0 +1,34 @@
...
+- Displaying to the user the funds allocated to them in the channel,
+- Getting approval from the user to proceed
+- Sending the withdraw + conclude to the blockchain
+- Getting acknowledgement from the user when done
+
+Out of scope (for the time being):
+
+- Retrying a transaction on failure.
+- Displaying an error message to the user on transaction failure.
+- Checking if a finalized outcome exists on chain(This will probably require a new saga).
```
tomclose: I was thinking about this because we need it for challenging too. It might make sense to have a section of the state where we just mirror the channel state from the adjudicator for channels we're interested in. That would mean storing the "funds held", "outcome" (if one exists) and "open challenge" and having a saga to keep them synced.

packages/wallet/src/redux/protocols/withdrawing/__tests__/scenarios.ts
```diff
@@ -0,0 +1,158 @@
...
+import * as testScenarios from '../../../__tests__/test-scenarios';
+import { EMPTY_OUTBOX_STATE } from '../../../outbox/state';
+import { Wallet } from 'ethers';
+
+// To test all paths through the state machine we will use 4 different scenarios:
+//
+// 1. Happy path: WaitForApproval -> WaitForTransaction -> WaitForAcknowledgement -> Success
+// 2. Withdrawal Rejected: WaitForApproval -> Failure
+// 3. Transaction failure: WaitForApproval -> WaitForTransaction -> Failure
+// 4. Channel not closed failure: WitForApproval -> Failure
```
tomclose: I'm beginning to think that the readme might be a good place to outline the scenarios?

packages/wallet/src/redux/protocols/transaction-submission/actions.ts
```diff
@@ -115,3 +115,16 @@ export const transactionFailed = (processId: string): TransactionFailed => ({
   type: TRANSACTION_FAILED,
   processId,
 });
+
+export const isTransactionAction = (action: BaseProcessAction): action is TransactionAction => {
```
tomclose: `BaseProcessAction` -> `WalletAction`?

packages/wallet/src/redux/protocols/withdrawing/states.ts
lalexgap: Oops

packages/wallet/src/redux/protocols/withdrawing/readme.md
lalexgap: I think that makes sense. Although practically I don't think we should ever let the user get there. We should only start the withdrawing protocol when we know the channel is closed.

tomclose: That's a good point. It kind depends how this protocol is going to be triggered, which isn't completely clear yet. Happy if you want to leave it out for now.

packages/wallet/src/redux/protocols/withdrawing/__tests__/reducer.test.ts
```diff
@@ -0,0 +1,113 @@
...
+  const { sharedData } = scenario;
+
+  describe('when initializing', () => {
+    const { processId, withdrawalAmount, channelId } = scenario;
+    const result = initialize(withdrawalAmount, channelId, processId, sharedData);
+
+    itTransitionsTo(result, states.WAIT_FOR_APPROVAL);
+  });
+
+  describe(whenIn(states.WAIT_FOR_APPROVAL), () => {
```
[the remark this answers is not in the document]
lalexgap: I think you're right. This was a pattern we introduced in some other tests with the idea that by using consistent wording we could generate a mermaid diagram from the tests. However I don't think it helps much right now.

packages/wallet/src/redux/protocols/withdrawing/readme.md
lalexgap: I think for now I'll put the channel closed check in initialize and dump into the failure state if its not. If we want to add a new acknowledge state that will be trivial to do in the future.

packages/wallet/src/redux/protocols/withdrawing/readme.md
lalexgap: That makes sense to me.

The other option of calling the contract when we need to check something seems slower and more painful to handle(since the calls will be async).

It sounds like we will need a saga that will:
 - take an adjudicator event
 - update channel state.
- possibly start a new process (IE: challenge detected would start the responding process)

Maybe we should create a card for this?

tomclose: https://github.com/magmo/apps/issues/384

Issue #369, opened 2019-04-16
Try out the scenarios testing approach for indirect funding protocol


kerzhner:
Both indirect funding unit tests and storybook tests must create mock states. `scenarios` will allow us to mock states just once.

PR #378, opened 2019-04-17
Indirect Funding: Test state stub sharing between unit tests and storybook

kerzhner:
Constructing states for testing is hard. Doing so once for unit tests and once for stories is twice as hard.

This PR moves the stub state construction to a scenario for indirect funding.

Fixes https://github.com/magmo/apps/issues/369.


packages/wallet/src/redux/protocols/indirect-funding/player-a/__tests__/stories.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import Modal from 'react-modal';
+import { Provider } from 'react-redux';
+import IndirectFundingContainer from '../../../../../containers/indirect-funding/indirect-funding';
```
tomclose: Would be nice to move the container into the `protocols/indirect-funding` folder.

packages/wallet/src/__stories__/b-dummy-wallet-states.ts
```diff
@@ -3,7 +3,7 @@ import * as fundingStates from '../redux/channel-state/funding/state';
 import * as sharedStates from '../redux/channel-state/shared/state';
```
tomclose: Is the plan to remove this file too at some point?

kerzhner: Yes, we should remove that file once indirect funding B player scenario is set up.

packages/wallet/src/redux/protocols/indirect-funding/player-a/__tests__/stories.tsx
kerzhner: Addressed in https://github.com/magmo/apps/pull/379.

PR #380, opened 2019-04-17
Responding Protocol

### LINKED ISSUE #356: Protocol for responding to a challenge
Write a protocol for responding to a challenge. To include:

* Deciding which options (respond/refute) are valid
* Getting confirmation from the user to launch a response
* Submitting the response transaction to the blockchain.
* Getting acknowledgement from the user before returning to app
* Returning to the app to make a move if required.

Out of scope (for now)

* Respond with alternative move

TODO:

- [ ] readme / state machine
- [ ] states
- [ ] actions
- [ ] test scenarios
- [ ] reducer
- [ ] reducer tests
- [ ] basic views
- [ ] stories

lalexgap:
Implements a basic responding protocol.
Includes:
- States/Actions
- Scenarios
- Reducer and tests
- Very basic views and stories

Out of scope:
Handling getting the response from  the application (or ledger channel)

Closes #356 .


packages/wallet/src/redux/protocols/responding/readme.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Open Questions
+
+- For a ledger channel we don't want to ask the user to select an option. We should be able to handle deciding that in the wallet. How do we handle that in the response protocol?
+  - Option 1: The responding protocol has another state (WaitForResponseStrategy) that determines if the user is asked to select an option.
+  - Option 2: Allow the responding protocol to have two start points (WaitForOptionSelection or WaitForApproval)
+  - Option 3: Separate responding into two different protocols. One that is responsible for submitting the provided response. One that is responsible for asking the user for a response.
+  - Option 4: Always ask the user?
+  - Option 5: Responding protocol always responds if possible based on existing commitment. Otherwise it waits for a response.
+- Should we bother letting the user know there's a difference between `RespondWithExistingMove` and `Refute`? Only one of those options should be valid at a time (I think). We could just prevent the user with the option of selecting a move or having the Wallet auto respond.
```
tomclose: I think we shouldn't bother. The rule should be "if you can refute, you refute. otherwise you respond".

lalexgap: Hmm based on this comment I've created a simplified state machine. 

Basically if we can respond to a challenge with the commitments on hand we do. Otherwise we wait for a response to be provided (assuming that there will be a protocol that handles this differently for app/ledger channels).

packages/wallet/src/redux/protocols/responding/__tests__/reducer.test.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  describe('when initializing', () => {
+    const { challengeCommitment } = scenario;
+    const result = initialize(processId, sharedData, challengeCommitment);
+
+    itTransitionsTo(result, states.WAIT_FOR_APPROVAL);
+    itSetsChallengeCommitment(result, scenario.challengeCommitment);
+  });
+
+  describe(`when in ${states.WAIT_FOR_APPROVAL}`, () => {
```
tomclose: I would probably go further and just have `describe('when in WaitForApproval', () => {`. I think that's marginally easier to read and you can then directly match with jest's output on the command line.

PR #381, opened 2019-04-17
Resolve issue 362

andrewgordstewart:
Right now, only one action is routed and only when it's player A.

The reason for restricting to player A is that the test scenarios aren't built out yet for player B, which I would prefer to do as a separate piece of work. At that point, both the code and the test can be copy-pasted from player A.


packages/wallet/src/redux/reducer.ts
```diff
@@ -63,9 +65,25 @@ function routeToProtocolReducer(state, action) {
...
-function routeToNewProcessInitializer(state, action) {
-  // TODO: Call the process creator
-  return state;
+function routeToNewProcessInitializer(
+  state: states.Initialized,
+  action: actions.protocol.NewProcessAction,
+) {
+  switch (action.type) {
+    case actions.indirectFunding.FUNDING_REQUESTED:
+      switch (action.playerIndex) {
```
andrewgordstewart: Actually, this logic should live in the `indirect-funding` protocol.

packages/wallet/src/redux/__tests__/initialized.test.ts
```diff
@@ -26,3 +31,12 @@ describe('when the player initializes a channel', () => {
...
   });
 });
+
+describe('when a NewProcessAction arrives', () => {
+  const action = actions.indirectFunding.fundingRequested(channelId, PlayerIndex.A);
+  const initialize = jest.fn();
+  Object.defineProperty(IndirectFunding, 'initialize', { value: initialize });
+
+  walletReducer(initializedState, action);
+  expect(initialize).toHaveBeenCalledWith(action, states.EMPTY_SHARED_DATA);
```
andrewgordstewart: I need to expect something about the updated wallet state's `processStore`

packages/wallet/src/redux/reducer.ts
```diff
@@ -63,9 +65,23 @@ function routeToProtocolReducer(state, action) {
...
+    case actions.indirectFunding.FUNDING_REQUESTED:
+      const { protocolState, sharedData } = indirectFunding.initialize(
+        action,
+        states.sharedData(state),
+      );
+
+      return startProcess(state, sharedData, action, protocolState);
+    default:
+      return state;
+    // TODO: Why is the discriminated union not working here?
```
lalexgap: Weird, what was the error from `unreachable` ?

packages/wallet/src/redux/reducer.ts
```diff
@@ -90,3 +106,25 @@ const waitForLoginReducer = (
       return state;
   }
 };
+function startProcess(
+  state: states.Initialized,
+  sharedData: states.SharedData,
+  action: {
```
lalexgap: Is there a reason we're not just using a `FundingRequested` action type?

packages/wallet/src/redux/reducer.ts
```diff
@@ -90,3 +106,25 @@ const waitForLoginReducer = (
...
+    type: 'WALLET.INDIRECT_FUNDING.FUNDING_REQUESTED';
+    channelId: string;
+    playerIndex: PlayerIndex;
+    protocol: WalletProtocol;
+  },
+  protocolState:
+    | states.indirectFunding.playerA.WaitForApproval
```
lalexgap: Could we type this more broadly? Like `ProtocolState`?

It seems strange to have a `startProcess` method expecting  just `IndirectFunding` arguments. It seems like it should take `ProtocolState` and a `ProtocolAction` even if thats only happens to be indirect funding stuff right now.

packages/wallet/src/redux/reducer.ts
andrewgordstewart:
```
Argument of type '{ type: "WALLET.INDIRECT_FUNDING.FUNDING_REQUESTED"; channelId: string; playerIndex: PlayerIndex; protocol: WalletProtocol; }' is not assignable to parameter of type 'never'.
```

?

packages/wallet/src/redux/reducer.ts
andrewgordstewart: Yeah, thanks for catching that. It's from a VSCode refactor that I neglected to update.

packages/wallet/src/redux/reducer.ts
tomclose: I've had this problem when there's only one type before. I think if there's only one possibility for a type, typescript doesn't bother narrowing it in the switch statement.

packages/wallet/src/redux/reducer.ts
```diff
@@ -90,3 +135,26 @@ const waitForLoginReducer = (
       return state;
   }
 };
+function startProcess(
+  state: states.Initialized,
+  sharedData: states.SharedData,
```
tomclose: Doesn't the `state` contain `sharedData`?

packages/wallet/src/redux/state.ts
```diff
@@ -71,27 +73,27 @@ export function indirectFundingOngoing(state: Initialized): state is IndirectFun
...
 // ------------
-export const emptyState: Shared = {
+export const EMPTY_SHARED_DATA: SharedData = {
   outboxState: emptyDisplayOutboxState(),
-  channelState: { initializedChannels: {}, initializingChannels: {} },
+  channelState: emptyChannelState(),
 };
 
-function shared(params: Shared): Shared {
+export function sharedData(params: SharedData): SharedData {
```
tomclose: Unrelated to this PR, but isn't this function pointless?

PR #383, opened 2019-04-17
Add funding readme

andrewgordstewart:


packages/wallet/src/redux/protocols/funding/readme.md
```diff
@@ -0,0 +1,59 @@
...
+
+  WFSC --> |Cancel| F((failure))
+  WFSR --> |Cancel| F
+```
+
+### Player B
+
+```mermaid
+graph TD
+  S((start)) --> WFSP(WaitForStrategyProposal)
```
andrewgordstewart: I think it makes more sense for player B's wallet to wait for the strategy proposal, then simply ask the app to approve or reject that strategy. If you
1. Wait for funding approval
2. Wait for strategy response
3. Wait for strategy approval

then the user has to approve twice, which is probably confusing.

That approach also adds complexity, since the strategy proposal can arrive before or after the funding approval. My proposed state machine is more linear.

packages/wallet/src/redux/protocols/funding/readme.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Player A
+
+```mermaid
+graph TD
+  S((start)) --> WFSC(WaitForStrategyChoice)
+
+  WFSC --> |StrategyChosen| WFSR(WaitForStrategyResponse)
+  WFSR --> |StrategyApproved| WFF(WaitForFunding)
+  WFF --> |StrategyAction| WFF
```
lalexgap: Minor: Would `FundingStrategyAction` be clearer? Just to differentiate it from `StrategyApproved`, `StrategyChosen`, etc..

PR #385, opened 2019-04-18
Challenging

tomclose:
Does most of the challenging protocol. Thought it would be a good time to merge though.

Still to do:

- [x] Basic views
- [x] Stories

### DIFF
--- a/packages/wallet/src/redux/__tests__/test-scenarios.ts
+++ b/packages/wallet/src/redux/__tests__/test-scenarios.ts
@@ -77,6 +77,15 @@ export const preFundCommitment2: Commitment = {
   allocation: twoThree,
   destination: participants,
 };
+export const gameCommitment0: Commitment = {
+  channel,
+  commitmentCount: 2,
+  commitmentType: CommitmentType.App,
+  appAttributes: '0x0',
+  turnNum: 18,
+  allocation: ['0x05', '0x05'],
+  destination: [],
+};
 export const gameCommitment1: Commitment = {
   channel,
   commitmentCount: 2,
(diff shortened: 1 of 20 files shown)


packages/wallet/src/redux/protocols/challenging/__tests__/reducer.test.ts
```diff
@@ -0,0 +1,214 @@
...
+import { FailureReason } from '../states';
+
+describe('opponent-responds scenario', () => {
+  const scenario = scenarios.opponentResponds;
+  const { channelId, processId, storage } = scenario;
+
+  describe('when initializing', () => {
+    const result = initialize(channelId, processId, storage);
+
+    itTransitionsTo(result, 'ApproveChallenge');
```
lalexgap: Can we make these type-safe in some way?

 Otherwise if we follow this pattern we're going to have to be very vigilant on typos and refactoring.

tomclose: Done with `export type ChallengingStateType = ChallengingState['type'];`.

(As it happens, I don't think the problem is actually that bad here anyhow, as any typo will be immediately caught by the test. The other case we use these strings is in discriminated unions and in direct comparisons with `state.type` - both of which will be picked up by the type checker.)

packages/wallet/src/redux/__tests__/test-scenarios.ts
```diff
@@ -86,6 +91,10 @@ export const gameCommitment1: Commitment = {
   allocation: ['0x05', '0x05'],
   destination: [],
 };
+export const signedCommitment19 = {
```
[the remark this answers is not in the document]
tomclose: I was actually thinking that this could be the new standard. It's useful to have the `turnNum` in the name and you can usually deduce the state type from it. (We might need to have `signedCommitment19c` for conclude states.) I will make further changes in a future PR.

PR #387, opened 2019-04-18
Store Adjudicator State in the store

### LINKED ISSUE #384: Adjudicator syncing
For several protocols (e.g. challenging and responding) it would be useful to have the state of the adjudicator in the redux store. The format would be something like:
```ts
interface Adjudicator {
    [channelId: string]: OnChainChannelState;
}

interface OnChainChannelState {
    balance: number; // or string or bigint or whatever
    outcome?: { destination: string, total: number }[];
    challenge?: { expiry: number, state: ChannelState };
}
```
The adjudicator syncer would be responsible for:

* Keeping the `Adjudicator` data up-to-date (for all `channelId`s in the `channelStore`)
* Triggering a `CHALLENGE_DETECTED` when a challenge is added to a channel.
* Triggering a `CHANNEL_FINALIZED` (a new action) when the outcome is finalized (either from a channel expiring or from a conclusion proof being presented).
* Triggering a `FUNDING_CHANGED` when a channel's balance changes
* Triggering an `OUTCOME_CHANGED` when the outcome changes (due to a `claim` or `transfer`)
* ... possibly some other things

Unanswered question: how do we route these actions to the right process? Do we just use the existing `channelId` subscription in the `Process` object?

lalexgap:
Adds `AdjudicatorState` and a reducer that handles adjudicator events and updates `AdjudicatorState`.

Also adds a challenge-saga (that replaces the block mining saga)  that emits a challenge expired event.

TODO: 

- [x] Add tests for new saga

Fixes #384


packages/wallet/src/redux/adjudicator-state/state.ts
```diff
@@ -0,0 +1,13 @@
...
+  [channelId: string]: AdjudicatorChannelState;
+}
+export interface AdjudicatorChannelState {
+  balance: string;
+  concluded: boolean;
+  challenge?: {
+    expiresAt: number;
+    challengeCommitment: Commitment;
+  };
+}
```
tomclose: Would be nice to have some helper functions here that could be used in the reducer e.g.

* For the `AdjudicatorState`:
    * `getChannel`
    * `setChannel`
* For the `AdjudicatorChannelState`:
    * `clearChallenge`
    * `setChallenge`
    * `setBalance`
    * `markAsFinalized`
    * etc.

Issue #366, opened 2019-04-15
Update the direct-funding protocol to use the transaction-submission protocol


tomclose:
Now that we have the transaction-submission protocol, we should be able to use this inside the direct-funding protocol.

We should also update the direct-funding protocol to follow the same format as the newer protocols:

TODO:
- [x] readme / state machine
- [x] states
- [x] actions
- [x] test scenarios
- [x] reducer
- [x] reducer tests
- [x] basic views
- [x] stories

PR #388, opened 2019-04-19
Direct funding uses transaction submission protocol

kerzhner:
This PR fixes https://github.com/magmo/apps/issues/366. With the inclusion of PostFund messages in the direct funding protocol, this PR is growing in scope. It is probably worth it to merge the PR in and continue the remaining work in a new PR.

The direct funding protocol states and transitions between those states are complete. Side effects that need to occur for actions are NOT complete. Specifically, the remaining work is:
- [ ] Action side effects: player B does not send post fund commitments at the moment. A set of `TODO`s marks where the commitments should be sent.
- [ ] Action side effects: channel state has to be updated to mirror the progression of the direct funding protocol.
- [ ] Handle non-happy paths such as transaction failure.


packages/wallet/src/redux/protocols/direct-funding/readme.md
```diff
@@ -0,0 +1,25 @@
...
+## State Machine
+
+The protocol implements the following state machine:
+
+TODO: add action names for transitions.
+
+```mermaid
+graph LR
+ST((Start))-->NSD(NotSafeToDeposit)
+NSD-->|FUNDING_RECEIVED|TSSM(TransactionSubmissionStateMachine)
```
tomclose: This should probably be the direct-funding state i.e. `WaitForDepositTransaction`.

packages/wallet/src/redux/protocols/direct-funding/readme.md
```diff
@@ -0,0 +1,25 @@
...
+ST-->TSSM
+TSSM-->WFFC(WaitForFundingConfirmation)
```
tomclose: This is probably already planned but wanted to leave a note that you'll need to handle the transaction-submission `Failure` case here too.

packages/wallet/src/redux/protocols/direct-funding/readme.md
```diff
@@ -0,0 +1,25 @@
+# Direct Funding Protocol
+
+The direct funding protocol is one of the funding protocols, which means that the protocol progresses the channel from after the exchange of the PreFund setup states to where the participants are able to exchange PostFund states. The keeps track of adjudicator funding in order to track when it is safe to deposit funds.
```
tomclose: I think it might make sense for the direct-funding-protocol to take care of the exchange of the post-fund states too. If you imagine you're the indirect funding protocol, you probably want this taken care of, so the case where you have to directly fund the ledger channel looks more similar to the case where you use an existing ledger channel.

andrewgordstewart: :+1:

It seems to me like any funding protocol, on its successful completion, should result in the target channel being in the application phase.

packages/wallet/src/redux/protocols/direct-funding/__tests__/scenarios.ts
```diff
@@ -0,0 +1,47 @@
...
+};
+
+const defaultsForB: states.DirectFundingState = {
+  ...defaultsForA,
+  requestedYourContribution: YOUR_DEPOSIT_B,
+  ourIndex: 1,
+  safeToDepositLevel: YOUR_DEPOSIT_A,
+};
+
+export const happyPathA = {
```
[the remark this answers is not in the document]
kerzhner: Addressed in https://github.com/magmo/apps/pull/388/commits/446b7d19cc2eb0467483c795955720bdd74241ac.

packages/wallet/src/redux/protocols/direct-funding/readme.md
kerzhner: Addressed in https://github.com/magmo/apps/pull/388/commits/446b7d19cc2eb0467483c795955720bdd74241ac.

packages/wallet/src/redux/protocols/direct-funding/readme.md
kerzhner: Agreed. In the short term, how should direct funding handle transaction failure? I think some of the options are:

1. Abandon direct funding by returning a terminal funding failure state.
2. Retry the transaction.

In the long term, direct funding should probably offer both options to the user.

tomclose: I think Option 1 should be ok for the time being.

packages/wallet/src/redux/protocols/direct-funding/readme.md
kerzhner: I am onboard with that. There is an open question about the implementation of the PostFund state exchange. For indirect funding, we chose to split the protocol into protocolA and protocolB. We could take the same approach for direct funding. Alternatively, we could try out a unified protocol approach.

andrewgordstewart: I'm personally for maintaining a single set of states. I'm currently writing the funding protocol, where the states and state machine is nearly identical, and it's tedious and distracting to always remember "Oh yeah, I should make this change in player-b states as well"

packages/wallet/src/redux/protocols/direct-funding/state.ts
```diff
@@ -1,111 +1,111 @@
...
-): WaitForFundingConfirmation {
+
+interface ConditionalParams {
+  channelFunded: boolean;
+  postFundSetupReceived: boolean;
+}
+export function waitForFundingAndPostFundSetup<T extends BaseDirectFundingState>(
+  params: Properties<BaseDirectFundingState>,
+  conditionalParams: ConditionalParams,
+): WaitForFundingAndPostFundSetup {
```
tomclose: A possible simplification here is to use `Properties` to pull in every property apart from `type` from the `WaitForFundingAndPostFundSetup`:
```ts
export function waitForFundingAndPostFundSetup(
    params: Properties<WaitForFundingAndPostFundSetup>
): WaitForFundingAndPostFundSetup {
    const { channelFunded, postFundSetupReceived } = params;
    return {
        ...baseDirectFundingState(params),
        type: WAIT_FOR_FUNDING_AND_POST_FUND_SETUP, 
        channelFunded,
        postFundSetupReceived,
    };
}
```
There's also now [a utility](https://github.com/magmo/apps/blob/master/packages/wallet/src/redux/utils.ts#L8-L16) to help with this pattern, allowing you to write it:
```ts
const waitForFundingAndPostFundSetup: Constructor<WaitForFundingAndPostFundSetup> = params => {
     // ....
}
```

kerzhner: Thanks for pointing this out. Addressed in https://github.com/magmo/apps/commit/3e7331214023bb2201c6aa913bfa5b45e09c1c27.

PR #390, opened 2019-04-19
Funding protocol work

andrewgordstewart:
* [x]  readme: state machine + sequence diagram
 * [x]  states
* [x]   actions
 * [x]  test scenarios
* [ ]   reducer
* [x]   reducer tests
* [ ]   basic views
* [ ]   stories

### DIFF
--- a/packages/wallet/src/redux/protocols/indirect-funding/actions.ts
+++ b/packages/wallet/src/redux/protocols/indirect-funding/actions.ts
@@ -1,7 +1,8 @@
 import * as playerA from './player-a/actions';
 import * as playerB from './player-b/actions';
-import { CommonAction } from '../../actions';
+import { CommonAction, WalletAction, isCommonAction } from '../../actions';
 import { PlayerIndex, WalletProtocol } from '../../types';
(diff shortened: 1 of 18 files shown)


packages/wallet/src/redux/protocols/indirect-funding/actions.ts
```diff
@@ -15,3 +16,13 @@ export type FundingRequested = ReturnType<typeof fundingRequested>;
...
 export type Action = FundingRequested | ProcessAction | CommonAction;
+
+export function isIndirectFundingAction(action: WalletAction): action is Action {
+  return isCommonAction(action)
+    ? true
+    : isDirectFundingAction(action)
+    ? true
+    : action.type.indexOf('WALLET.INDIRECT_FUNDING') === 0
+    ? true
+    : false;
```
tomclose: This is a kind-of mind-boggling way to write this. Maybe something like the following would be clearer:
```ts
export function isIndirectFundingAction(action: WalletAction): action is Action {
    return isCommonAction(action) || isDirectFundingAction(action) || action.type.indexOf('WALLET.INDIRECT_FUNDING') === 0;
}
```

PR #391, opened 2019-04-21
use function modifiers in the RPS contract

geoknee:
This should mean less code duplication, and make it slightly easier to audit the contract. It was also an excuse to revisit the contract tests: I added one for resolutions being changed incorrectly. I was about to do the same for the 'integer overflow prevented' check, but realised that I did not understand why it has the name it does.

It seems clear that both players should have enough funds to pay the stake to the winner if they lose -- but the message in the relevant `require` statement wasn't doing much justice to that. 

Can anyone enlighten me about the overflow? If a user's allocation has overflowed, it seems to me that we have far bigger problems than safely starting a new game...

### DIFF
--- a/packages/rps/contracts/RockPaperScissorsGame.sol
+++ b/packages/rps/contracts/RockPaperScissorsGame.sol
@@ -63,14 +63,14 @@ contract RockPaperScissorsGame {
         }
     }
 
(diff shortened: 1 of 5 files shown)


geoknee:
Thanks @andrewgordstewart! The underflow thing makes sense, I will take a look at using `SafeMath` as a part of this PR.

geoknee:
I have one more commit that I would like to squeeze in, where I have gotten rid of the reference to overflow.

PR #394, opened 2019-04-22
Concluding protocol

### LINKED ISSUE #359: Protocol for concluding
Write a protocol for ~~resigning from~~ concluding a channel (i.e. moving to conclude). To include:

* Checking to see if it is the player's turn and explaining they can't resign if not
* Asking user to confirm the resignation (probably displaying the current outcome)
* Formulating the conclude state and sending to the opponent
* Waiting for a conclude from the opponent
* Acknowledge channel closed (giving the option to defund)

Out of scope (for now):

* Giving the option to launch a challenge if the conclude doesn't arrive

TODO:
- [x] readme / state machine
- [x] states
- [x] actions
- [x] test scenarios
- [x] reducer tests (w/o side effects)
- [x] reducer (w/o side effects)
- [x] basic views
- [x] stories
- [ ] finalize scope of this protocol
- [ ] reducer tests (w side effects)
- [ ] reducer (w side effects)

geoknee:
Will close #359.

- [x] readme / state machine
- [x] states
- [x] actions
- [x] test scenarios
- [x] reducer tests (w/o side effects)
- [x] reducer (w/o side effects)
- [x] basic views
- [x] stories
- [x] finalize scope of this protocol
- [ ] reducer tests (w side effects)
- [ ] reducer (w side effects)


packages/wallet/src/redux/protocols/resigning/readme.md
```diff
@@ -0,0 +1,40 @@
...
+- Giving the option to launch a challenge if the conclude doesn't arrive
+
+## State machine
+
+The protocol is implemented with the following state machine
+
+```mermaid
+graph TD
+  S((start)) --> CR{My turn?}
+  CR  --> |Yes| CC(ConfirmResignation)
```
tomclose: This should probably be `ApproveResignation` to be consistent with the other protocols.

packages/wallet/src/redux/protocols/resigning/readme.md
```diff
@@ -0,0 +1,40 @@
...
+  CR  --> |No| RC(ResignationImpossible)
```
tomclose: `AcknowledgeResignationImpossible`?

packages/wallet/src/redux/protocols/resigning/readme.md
```diff
@@ -0,0 +1,40 @@
...
+  CC  --> |CONCLUDE.SENT| WOC(WaitForOpponentConclude)
+  WOC --> |CONCLUDE.RECEIVED| ACC(AcknowledgeChannelClosed)
+  ACC --> |DEFUND.CHOSEN| D(Defund)
+  D   --> |DEFUNDED| SS
+  ACC --> |DEFUND.NOT.CHOSEN| SS((success))
+  RC  --> |ACKNOWLEDGE.RESIGNATION.IMPOSSIBLE| F((failure))
```
tomclose: `ResignationImpossibleAcknowledged`? (On the whole actions are -ed words.)

packages/wallet/src/redux/protocols/resigning/readme.md
tomclose: This could even be `AcknowledgeFailure`

packages/wallet/src/redux/protocols/resigning/readme.md
```diff
@@ -0,0 +1,40 @@
...
+The protocol is implemented with the following state machine
+
+```mermaid
+graph TD
+  S((start)) --> CR{My turn?}
+  CR  --> |Yes| CC(ConfirmResignation)
+  CR  --> |No| RC(ResignationImpossible)
+  CC  --> |CONCLUDE.SENT| WOC(WaitForOpponentConclude)
+  WOC --> |CONCLUDE.RECEIVED| ACC(AcknowledgeChannelClosed)
+  ACC --> |DEFUND.CHOSEN| D(Defund)
```
tomclose: `WaitForDefund`?

packages/wallet/src/redux/protocols/resigning/states.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export interface WaitForOpponentConclude {
+    type: 'WaitForOpponentConclude';
+    processId: string;
+}
+
+export interface AcknowledgeChannelClosed {
+    type: 'AcknowledgeChannelClosed';
+    processId: string;
+}
```
tomclose: Looks like you're missing the `AcknowledgeFailure` state here (or whatever you decide to call it).

geoknee: 👍 thanks

geoknee:
1. Should we have a fork in the current happy path for Defund Success vs Defund Failure? Or is that out of scope for now?

2. Is it likely that users will want to resign without defunding the channel? We could bump DEFUND.NOT.CHOSEN out of scope for now, for simplicity? I suppose if I have been absolutely beaten and lost all of my money to the other player, I wouldn't bother defunding.

tomclose:
1. Probably yes. You should handle all of the possible terminal states of any sub-protocol that you call.

2. Yeah, we could probably get rid of this.

geoknee:
1. Would it make more sense to have `Defund` initialised *after* `Resigning`, and by the parent protocol of `Resigning`? If the user gets (via Happy Path) to `AcknowledgeChannelClosed`, but then gets `DEFUND.FAILED`, it is not clear whether this is an overall success or failure for the Resigning Protocol. The two protocols seem to be fairly well independent, and the parent process can decide when to call each (most likely `Resigning` and then `Defunding`)

2. Ok

packages/wallet/src/redux/protocols/concluding/readme.md
```diff
@@ -0,0 +1,61 @@
...
+  D   --> |DEFUND.FAILED| F3((Failure))
+  RC  --> |CONCLUDING.IMPOSSIBLE.ACKNOWLEDGED| F4((Failure))
+  style S  fill:#efdd20
+  style E  fill:#efdd20
+  style MT fill:#efdd20
+  style SS fill:#58ef21
+  style F1 fill:#f45941
+  style F2 fill:#f45941
+  style F3 fill:#f45941
+  style F4 fill:#f45941
```
tomclose: Love the colours!

packages/wallet/src/redux/protocols/concluding/readme.md
```diff
@@ -0,0 +1,61 @@
...
+- Allowing the user to not choose to not defund
+
+## State machine
+
+The protocol is implemented with the following state machine
+
+```mermaid
+graph TD
+  S((Start)) --> E{Channel Exists?}
+  E --> |No| F1((Failure))
```
tomclose: If this is going to be a top-level protocol (which it looks like it will be) you probabl need an `AcknowledgeFailure` here and before all transitions into the `Failure` state.

packages/wallet/src/redux/protocols/concluding/states.ts
```diff
@@ -0,0 +1,109 @@
...
+// Constructors
+// ------------
+export function acknowledgeConcludingImpossible(
+  p: P<AcknowledgeConcludingImpossible>,
+): AcknowledgeConcludingImpossible {
+  const { processId } = p;
+  return { type: 'AcknowledgeConcludingImpossible', processId };
+}
+
+export function approveConcluding(p: P<ApproveConcluding>): ApproveConcluding {
```
tomclose: Could use the `Constructor` helper here instead of `P`.

packages/wallet/src/redux/protocols/concluding/readme.md
geoknee: I added two such states (I don't think more are necessary, apart from perhaps one before transition into the Success state).

packages/wallet/src/redux/protocols/concluding/states.ts
```diff
@@ -37,6 +47,16 @@ export interface AcknowledgeChannelConcluded {
...
 
+export interface AcknowledgeChannelDoesntExist {
+  type: 'AcknowledgeChannelDoesntExist';
+  processId: string;
+}
+
+export interface AcknowledgeDefundFailed {
+  type: 'AcknowledgeDefundFailed';
+  processId: string;
+}
```
tomclose: Another option here would be to have:
```ts
export interface AcknolwedgeFailure {
    type: 'AcknowledgeFailure',
    processId: string,
    failure: FailureReason,
}
```
It really depends how similarly you treat these states throughout the protocol.

geoknee: 👍 I may make this optimisation in a future PR.

packages/wallet/src/redux/protocols/concluding/states.ts
geoknee: Done.

PR #396, opened 2019-04-22
De-Funding Protocol

lalexgap:
A protocol responsible for defunding a channel. Based on the channel type it delegates to either the withdrawal protocol or ledger de-funding protocol.

Once the the ledger de-funding protocol is completed this protocol might need to be revisited and slightly tweaked.

No stories were created as there are no unique views for the protocol

Fixes #396

### DIFF
--- a/packages/wallet/src/redux/state.ts
+++ b/packages/wallet/src/redux/state.ts
@@ -69,7 +69,7 @@ export interface ProcessState {
 }
 
 export interface FundingState {
-  [channelId: string]: FundingState;
+  [channelId: string]: ChannelFundingState;
 }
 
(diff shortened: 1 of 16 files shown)


packages/wallet/src/redux/state.ts
```diff
@@ -67,6 +68,15 @@ export interface ProcessState {
   channelsToMonitor: string[];
 }
 
+export interface FundingState {
```
lalexgap: I think we'll need some permanent store of whether a channel is funded directly or indirectly and that stores associations between ledger channels and the channels they fund. I'm not sure if this should be in `ChannelState` or if we should have a `FundingState`.

PR #397, opened 2019-04-23
Ledger De-Funding

lalexgap:
Indirect de-funding reducer, scenarios and tests. This assumes that player A will always go first. There is an open issue #412 to address this.

The views and stories will be handled in #411.

Fixes #397

### DIFF
--- a/packages/wallet/src/redux/protocols/indirect-defunding/readme.md
+++ b/packages/wallet/src/redux/protocols/indirect-defunding/readme.md
@@ -14,24 +14,21 @@ It covers:
 
 ```mermaid
 graph TD
-  St((start))-->WFAp(WaitForApproval)
-  WFAp-->|Approve|SLU[SendLedgerUpdate]
-  WFAp-->|Rejected|F((failure))
-  SLU-->WFL(WaitForLedgerUpdate)
-  WFL-->|LedgerUpdateReceived|WFAc(WaitForAcknowledgement)
-  WFAc-->|Acknowledged|S((success))
+  St((start))-->DF{Defundable?}
+  DF --> |No| F((Failure))
+  DF --> |Yes| WFU(WaitForLedgerUpdate1)
+  WFU --> |"CommitmentReceived(Accept)"| Su((Success))
+  WFU --> |"CommitmentReceived(Reject)"| F
 ```
 
 ### Player B State machine
 
 ```mermaid
 graph TD
-  St((start))-->WFAp(WaitForApproval)
-  WFAp-->|Approve|WFL(WaitForLedgerUpdate)
-  WFAp-->|Rejected|F((failure))
-  WFL-->SLU[SendLedgerUpdate]
-  SLU-->WFAc(WaitForAcknowledgement)
-  WFAc-->|Acknowledged|S((success))
+  St((start))-->DF{Defundable?}
+  DF --> |No| F((Failure))
+  DF --> |Yes| WFU(WaitForLedgerUpdate0)
+  WFU --> |CommitmentReceived| Su((Success))
 ```
 
 Notes:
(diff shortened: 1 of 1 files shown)


packages/wallet/src/redux/protocols/indirect-defunding/readme.md
```diff
@@ -0,0 +1,44 @@
...
+  WFAc-->|Acknowledged|S((success))
+```
+
+Notes:
+
+- Currently doesn't handle the case where one player approves, one player declines
+
+## Open Questions
+
+1. Is this a top-level protocol? If so what closes the channel?
```
lalexgap: Are we expecting the ledger de-funding protocol to be called from a closing protocol of some kind? Right now I don't think we have a protocol for closing the channel. (Previously we had a closing reducer in channel state that handled exchanging the conclude proof).

If so I think the ledger de-funding can be pretty simple. We can assume that a higher level protocol has already 
done the "handshake" so we can assume the ledger de-funding protocol is co-operative and probably don't have to ask for approval.

It seems like we need a higher level protocol to "wrap up" an application. This would be responsible for letting the other wallet know that we want to conclude a game (either by sending a message or a conclude commitment) and triggering the ledger de-funding and then withdrawal protocol. I'm not sure if this is out of scope for this sprint or we thought that's the role the ledger de-funding protocol would play?

tomclose: Yes - I think this will be called from the defunding protocol.

packages/wallet/src/redux/protocols/indirect-defunding/readme.md
```diff
@@ -0,0 +1,44 @@
...
+- Crafting a ledger update that allocates the funds to the players.
+- Waiting for a ledger response from the opponent.
+
+## State machine
+
+### Player A State machine
+
+```mermaid
+graph TD
+  St((start))-->WFAp(WaitForApproval)
```
tomclose: Probably don't need this, as we can assume that approval has already been given.

packages/wallet/src/redux/protocols/indirect-defunding/readme.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+Notes:
+
+- SendLedgerUpdate0/1 are not states but indicate when the ledger update is sent.
+- A single reducer implements both the player A and B state machine.
+
+Assumptions:
+
+- We assume that Player A was responsible for the original set of ledger updates to fund the channel. Due to the current consensus algorithm if player A proposed a consensus (to fund the channel), player B must propose the next consensus (to de-fund the channel).
```
lalexgap: I only recently realized that Player B has to be responsible for sending the first ledger update to defund the channel. I've updated the readme but I'll need to go through and swap most of the test cases / player checks.

Longer term this might cause us headaches. We'll need to keep track of who was responsible for the last consensus round and have the next player start that consensus. For now I'm considering that out of scope.

andrewgordstewart: The simplest thing to do is for everyone but player `0` to respond to any ledger updates, either by reseting the consensus counter or agreeing with the proposed allocation.

This always leaves player `0` (player A) able to propose a ledger update.

lalexgap: Isn't Player A responsible for sending the commitment that sets allocation=proposedAllocation? Meaning after all other players have accepted the consensus Player A must send out a commitment that has `allocation`=`proposedAllocation` ?

andrewgordstewart: yes, but the rest of the players can continue to increment the consensus counter, giving player A the opportunity to change `proposedAllocation`

lalexgap: Ah, I think I understand. I'm going to leave Player A as the first mover for now and we can discuss next standup if we want to either:
- Change Indirect Funding so that Player B sends another ledger update
- Change Indirect Defunding so that player B is the first mover.

PR #398, opened 2019-04-23
Update transaction submission

tomclose:
Makes a few improvements to the transaction-submission protocol, updating it to the style used in later protocols.


geoknee:
Also, +1 for 'cruft'. I had to look that one up 😄

tomclose:
1. I wanted to try out the `Constructor` thing for a bit before making the call on whether we want to use it everywhere. If other people start to use it, I'd be happy to move it up then.
2. I did that :) : https://github.com/magmo/apps/pull/398/files#diff-46ef22ebaeddc66a599efcc6fdca752f

geoknee:
1. Fair enough. Moving it up =/= deciding to use it everywhere, though. If its flagged as a utility then it could be considered capable/staged for use elsewhere. 

2. My bad! Maybe I have red/green colorblindness

Issue #399, opened 2019-04-23
Add PostFund messages to Direct Funding


kerzhner:
Follow up work from https://github.com/magmo/apps/pull/388.

The direct funding protocol states and transitions between those states are complete. Side effects that need to occur for actions are NOT complete. Specifically, the remaining work is:

Action side effects: player B does not send post fund commitments at the moment. A set of TODOs marks where the commitments should be sent.
- [x] Address PR feedback: https://github.com/magmo/apps/pull/388#discussion_r277639861
- [x] Action side effects: player B does not send post fund commitments at the moment. A set of TODOs marks where the commitments should be sent.
- [x] Action side effects: channel state has to be updated to mirror the progression of the direct funding protocol.
- [ ] Handle non-happy paths such as transaction failure.

PR #405, opened 2019-04-23
Side effects for direct funding

kerzhner:
Fixes https://github.com/magmo/apps/issues/399. The changes are:
1. Todo comments are added where channel state must be updated. 
1. Direct funding protocol state stores `processId`.
1. `Constructor` generic type is used for state constructors. 
1. Common reducer utilities factored out to `reducer-helpers.ts`

I made a decision to not wire up the direct funding protocol to the current channel state reducer. Keeping the channel state in sync with the direct funding protocol is clunky at the moment, and will have to be redone once the channel state is simplified.

https://zube.io/magmo/apps/c/813 and https://zube.io/magmo/apps/c/812 have been created to track follow up work.


packages/wallet/src/__stories__/dummy-wallet-states.ts
```diff
@@ -6,8 +6,8 @@ const {
   channelNonce,
   libraryAddress,
   participants,
-  preFundCommitment1,
-  preFundCommitment2,
+  preFundCommitment0: preFundCommitment1,
+  preFundCommitment1: preFundCommitment2,
```
tomclose: I'm guessing that you didn't mean to do this here?

kerzhner: This is the result of symbol renaming mock `preFundCommitment1` and `preFundCommitment2` to `preFundCommitment0` and `preFundCommitment1` respectively. So this was intentional. Though the right approach in this particular file is to rename the local variables as well.

PR #413, opened 2019-04-24
Player A sends funding strategy to player B

andrewgordstewart:
Two choices were made here:
A. the opponent's messaging address is stored directly on the protocol state, as an alternative to extracting the participant's _signing_ address from the channel store
B. player A sends an action to player B, so that player B's top level reducer can, if they wish, extract the action from the message and route it to a process' reducer.

I'd like feedback on these two choices before moving forward.

### DIFF
--- a/packages/wallet/src/redux/protocols/funding/player-a/__tests__/scenarios.ts
+++ b/packages/wallet/src/redux/protocols/funding/player-a/__tests__/scenarios.ts
@@ -22,12 +22,15 @@ import { EMPTY_SHARED_DATA } from '../../../../state';
 // ---------
 const processId = 'process-id.123';
 const sharedData = EMPTY_SHARED_DATA;
+const targetChannelId = '0x123';
+const opponentAddress = '0xf00';
 
 const props = {
   processId,
   sharedData,
   fundingState: 'funding state' as 'funding state',
-  targetChannelId: '0x123',
+  targetChannelId,
+  opponentAddress,
 };
 
 // ----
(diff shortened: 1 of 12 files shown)


packages/wallet/src/redux/protocols/funding/player-a/reducer.ts
```diff
@@ -53,7 +54,15 @@ function strategyChosen(
   if (state.type !== states.WAIT_FOR_STRATEGY_CHOICE) {
     return { protocolState: state, sharedData };
   }
-  return { protocolState: states.waitForStrategyResponse(state), sharedData };
+  const { processId, opponentAddress } = state;
+  const { strategy } = action;
+  const sentAction = strategyProposed(processId, strategy);
+  const payload = { processId, data: { sentAction } };
+  const message = messageRelayRequested(opponentAddress, payload);
```
lalexgap: As we discussed I think if `handleIncomingMessage` can transform the action for Player B this could be a nice pattern to follow. I'm in favor of B.

packages/wallet/src/redux/protocols/funding/player-a/__tests__/scenarios.ts
```diff
@@ -22,12 +23,17 @@ import { EMPTY_SHARED_DATA } from '../../../../state';
...
+const targetChannelId = '0x123';
+const opponentAddress = '0xf00';
 
 const props = {
   processId,
   sharedData,
   fundingState: 'funding state' as 'funding state',
-  targetChannelId: '0x123',
+  targetChannelId,
+  opponentAddress,
```
lalexgap: I think this makes sense to do at some point but I'm wondering if we should hold off and do this consistently through our protocol reducers?

PR #422, opened 2019-04-25
Simplify channel

tomclose:
Begin to strip out parts of the channel-store.


packages/wallet/src/redux/channel-store/channel-state/valid-transition.ts
```diff
@@ -0,0 +1,13 @@
+import { ChannelState } from './states';
+import { Commitment, getChannelId } from '../../../domain';
+
+export function validTransition(state: ChannelState, commitment: Commitment): boolean {
```
geoknee: Could you help me understand why this function accepts two different types?

tomclose: It doesn’t have to. It’s just mostly used that way.

PR #423, opened 2019-04-26
Finish off Concluding protocol

### LINKED ISSUE #407: Finish off Concluding protocol
Conclude commitment
- [ ] actually craft and send
- [ ] test this^
Reference: `channel-states/closing/*`

`Defund` sub-protocol
- [x] Actually spool up the protocol at the appropriate point
- [x] Parent reducer needs to handle actions of child protocol
- [x] `DEFUND_SUCCEEDED` and `DEFUND_FAILED` should not be actions
Reference: `challenging` protocol

Cleaning up
- [x] Unify the various 'Acknowledge' states
- [x] Furnish with data that is displayed to user
- [x] Differentiate actions from non-actions in readme diagram (caps/non-caps?)

They conclude, not me
- [ ] Consider new protocol for this
Reference: `channel-states/closing/*`

geoknee:
Will close #407.

Conclude commitment
- [x] actually craft and send
- [x] test this^
Reference: `channel-states/closing/*`

`Defund` sub-protocol
- [x] Actually spool up the protocol at the appropriate point
- [x] Parent reducer needs to handle actions of child protocol
- [x] `DEFUND_SUCCEEDED` and `DEFUND_FAILED` should not be actions
Reference: `challenging` protocol
- [x] fixup scenarios, tests
- [x] fix final test, consider storing `channelId` on certain concluding states.

Cleaning up
- [x] Unify the various 'Acknowledge' states
- [x] Furnish with data that is displayed to user
- [x] Differentiate actions from non-actions in readme diagram (caps/non-caps?)

They conclude, not me
- [ ] Consider new protocol for this
Reference: `channel-states/closing/*`

### DIFF
--- a/packages/wallet/src/redux/protocols/concluding/__tests__/reducer.test.ts
+++ b/packages/wallet/src/redux/protocols/concluding/__tests__/reducer.test.ts
@@ -24,11 +24,11 @@ describe('[ Happy path ] scenario', () => {
     const action = scenario.actions.concludeReceived;
     const result = concludingReducer(state, storage, action);
 
-    itTransitionsTo(result, 'AcknowledgeChannelConcluded');
+    itTransitionsTo(result, 'AcknowledgeConcludeReceived');
   });
 
-  describe('when in AcknowledgeChannelConcluded', () => {
-    const state = scenario.states.acknowledgeChannelConcluded;
+  describe('when in AcknowledgeConcludeReceived', () => {
+    const state = scenario.states.acknowledgeConcludeReceived;
     const action = scenario.actions.defundChosen;
     const result = concludingReducer(state, storage, action);
 
@@ -37,6 +37,14 @@ describe('[ Happy path ] scenario', () => {
 
   describe('when in WaitForDefund', () => {
     const state = scenario.states.waitForDefund;
+    const action = scenario.actions.defunded;
+    const result = concludingReducer(state, storage, action);
+
+    itTransitionsTo(result, 'AcknowledgeSuccess');
+  });
+
+  describe('when in AcknowledgeSuccess', () => {
+    const state = scenario.states.acknowledgeSuccess;
     const action = scenario.actions.defunded;
     const result = concludingReducer(state, storage, action);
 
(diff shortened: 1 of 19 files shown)


packages/wallet/src/redux/protocols/concluding/__tests__/reducer.test.ts
```diff
@@ -15,7 +15,7 @@ describe('[ Happy path ] scenario', () => {
     const state = scenario.states.approveConcluding;
     const action = scenario.actions.concludeSent;
     const result = concludingReducer(state, storage, action);
-    // TODO check that the conclude has actually been sent
+    expect(result.sideEffects.messageOutbox.messagePayload.data.concludeCommitment).toBeDefined();
```
geoknee: I am not sure that this is sufficient for the commitment to actually get sent?

packages/wallet/src/redux/protocols/concluding/__tests__/reducer.test.ts
```diff
@@ -24,11 +24,11 @@ describe('[ Happy path ] scenario', () => {
     const action = scenario.actions.concludeReceived;
```
geoknee: This action needs to be wired up to somewhere else in the app (the message listener saga?)

geoknee:
I think this work could be merged in, with the understanding that the final work:

- sending/receiving commitments properly wired up
- other player resigning is dealt with in a separate protocol / by extending this protocol

as a part of #431.

packages/wallet/src/redux/protocols/concluding/__tests__/reducer.test.ts
andrewgordstewart: You should use `itSendsThisMessage`, so that this side effect is tested in a consistent manner.

geoknee: I used `expectThisCommitmentSent` (defined in the same helper file `src/redux/__tests__/helpers.ts` ) as that seemed more specific.

PR #389, opened 2019-04-19
Rework indirect-funding protocol

tomclose:

PR #427, opened 2019-04-28
Reworking indirect funding

tomclose:
This PR adds back in (and updates) a lot of the indirect funding work that was removed with the channel simplification. It replaces https://github.com/magmo/apps/pull/389.

Still to do (in a separate piece of work):

- [ ] Write reducer code to make the tests pass.


packages/wallet/src/redux/protocols/indirect-funding/player-a/__tests__/reducer.test.ts
```diff
@@ -0,0 +1,86 @@
...
+// -------
+type ReturnVal = ProtocolStateWithSharedData<IndirectFundingState>;
+
+function itTransitionsTo(state: ReturnVal, type: IndirectFundingState['type']) {
+  it(`transitions protocol state to ${type}`, () => {
+    expect(state.protocolState.type).toEqual(type);
+  });
+}
+
+function itSendsMessage(state: ReturnVal, message: SignedCommitment) {
```
andrewgordstewart: There's currently an `itSendsThisMessage` test helper, which could be used in place of this. I'm in favour of using a consistent test helper to expect side effects, rather than having each test suite define a bespoke helper.

The existing helper could be improved to meet the use case here (and its use updated in just a couple of places)

packages/wallet/src/redux/protocols/indirect-funding/player-b/reducer.ts
```diff
@@ -1,10 +1,219 @@
...
+  const unchangedState = { protocolState, sharedData };
+  if (action.type !== actions.COMMITMENT_RECEIVED) {
+    throw new Error('Incorrect action');
+  }
+
+  const checkResult = checkAndStore(sharedData, action.signedCommitment);
+  if (!checkResult.isSuccess) {
+    throw new Error('Indirect funding protocol, unable to validate or store commitment');
+  }
+  sharedData = checkResult.store;
```
lalexgap: Isn't this mutating the existing state instead of creating a new copy (which can create a lot of headaches)?

lalexgap: Nevermind, that doesn't seem to be the case.

PR #428, opened 2019-04-28
Application protocol

### LINKED ISSUE #429: Application protocol
Create a protocol whose responsibility is to listen for application commitments and update the channel store.

andrewgordstewart:
Add a new application protocol that is responsible for handling commitments received from the application.

It has two main responsibilities:
- Handling a request to generate an address private key.
- Handling requests to sign and validate commitments from the application.

Currently it uses a fix process id as we only expect one app channel to be running at a time.

Resolves #429.


packages/wallet/src/redux/channel-store/reducer.ts
```diff
@@ -2,21 +2,38 @@ import { ChannelStore, getChannel, setChannel } from './state';
...
+import * as channelActions from './actions';
 
-export const channelStateReducer: ReducerWithSideEffects<ChannelStore> = (
+export const channelStoreReducer: ReducerWithSideEffects<ChannelStore> = (
   state: ChannelStore,
-  action: WalletAction,
+  action: channelActions.ChannelAction,
 ): StateWithSideEffects<ChannelStore> => {
+  switch (action.type) {
+    case channelActions.OPPONENT_COMMITMENT_RECEIVED:
```
lalexgap: This was added to support existing protocols that rely on the channel state reducer. If we switch everything to `signAndStore`/`checkAndStore` these can go,

PR #439, opened 2019-04-30
Changes to get Funding Process Working

lalexgap:
- Changed some plumbing so the app protocol works correctly
- Add basic ProtocolContainer so we can see the funding views again


kerzhner:
This looks good to me.

PR #440, opened 2019-04-30
Update shared data before protocol reducers

### LINKED ISSUE #435: Decide how the relationship between the adjudicator-watcher and the reducers should work
* Events come in from the chain
* How do we make sure that the adjudicator state is updated before protocols are pinged?

lalexgap:
Added some simple logic so that we update the shared data before calling the protocol reducers.

Right now this only handles `AdjudicatorEventAction`s.

Fixes #435

### DIFF
--- a/packages/wallet/src/redux/actions.ts
+++ b/packages/wallet/src/redux/actions.ts
@@ -193,3 +193,20 @@ export {
   isCommonAction,
   application,
 };
+
+// These are any actions that update shared data directly without any protocol
+export type SharedDataUpdateAction = AdjudicatorEventAction;
+
+export function isSharedDataUpdateAction(action: WalletAction): action is SharedDataUpdateAction {
+  return isAdjudicatorEventAction(action);
+}
+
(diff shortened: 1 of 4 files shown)


packages/wallet/src/redux/actions.ts
```diff
@@ -193,3 +193,20 @@ export {
...
+// These are any actions that update shared data directly without any protocol
+export type SharedDataUpdateAction = AdjudicatorEventAction;
+
+export function isSharedDataUpdateAction(action: WalletAction): action is SharedDataUpdateAction {
+  return isAdjudicatorEventAction(action);
+}
+
+export function isAdjudicatorEventAction(action: WalletAction): action is AdjudicatorEventAction {
+  return (
+    action.type === CONCLUDED_EVENT ||
```
kerzhner: We could filter based on the `WALLET.ADJUDICATOR` prefix. This approach would allows us to add new adjudicator events without updating the filter.

PR #441, opened 2019-04-30
Indirect Funding Work

lalexgap:
Work done on indirect funding.

Here's a brief summary of the changes:
- Changed Player B reducer to call `checkAndInitialize` to initialize the state in the `waitForPrefundSetup`
- Finished up the `handleWaitForPostFundSetup1` in player B reducer.
- Updated player B reducer to update funding state on `protocolState` when in the `waitForDirectFunding` state. (Might be worth creating a test that would catch this)
- Various other changes in the player B reducer to get tests passing.
- Start on player A reducer to get some of the tests passing. `Initialize`, `handlePreFundSetup1`, and `handleWaitForDirectFunding` are now written and passing tests.

### DIFF
--- a/packages/wallet/src/domain/commitments/index.ts
+++ b/packages/wallet/src/domain/commitments/index.ts
@@ -94,7 +94,7 @@ export function nextSetupCommitment(commitment: Commitment): Commitment | 'NotAS
   if (turnNum < numParticipants) {
     commitmentType = CommitmentType.PreFundSetup;
     commitmentCount = turnNum;
-  } else if (turnNum < 2 * numParticipants - 1) {
+  } else if (turnNum <= 2 * numParticipants - 1) {
     commitmentType = CommitmentType.PostFundSetup;
     commitmentCount = turnNum - numParticipants;
   } else {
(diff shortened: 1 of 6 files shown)


packages/wallet/src/domain/commitments/index.ts
```diff
@@ -45,6 +46,46 @@ export function constructConclude(commitment: Commitment): Commitment {
...
+  if (commitment.commitmentType !== CommitmentType.App || turnNum <= 2 * numParticipants) {
+    return 'NotAnUpdateCommitment';
+  }
+
+  let allocation;
+  let destination;
+  let updatedAppAttributes;
+
+  // This is the final commitment so we must reset to 0 and update allocation/destination
+  if (turnNum === 3 * numParticipants - 1) {
```
geoknee: Should this not happen on, e.g. `turnNum === 3` ?

Assuming `numParticipants = 2`, the RHS of this condition evaluates to 5.

packages/wallet/src/domain/commitments/index.ts
```diff
@@ -53,7 +94,7 @@ export function nextSetupCommitment(commitment: Commitment): Commitment | 'NotAS
   if (turnNum < numParticipants) {
     commitmentType = CommitmentType.PreFundSetup;
     commitmentCount = turnNum;
-  } else if (turnNum < 2 * numParticipants - 1) {
+  } else if (turnNum <= 2 * numParticipants - 1) {
```
geoknee: Should this not be 

`} else if (turnNum <= 2 * numParticipants ) {`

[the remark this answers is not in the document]
geoknee: Reading this function more carefully, these conditions are looking at the `turnNum` of the updated (returned) commitment, not the incoming (accepted) commitment.  So the `<=` is correct.

PR #442, opened 2019-05-01
Indirect funding with the new Consensus App

geoknee:
The consensus app now includes an `updateType` field and helper functions such as `vote`, `propose` and so on. This PR looks to update the tests and reducers for the `indirect-funding` protocol, such that they will craft and send appropriate ledger channel commitments using the new helpers. 

There is an outstanding issue about where the source of truth should be for the consensus app. The contracts themselves live in a separate package in a repository (soon to be moved to a separate package in this repository). The same is true for some typescript helpers. Other typescript helpers live in `apps/packages/wallet/src/domain/two-player-consensus-game/index.ts` as well as in `apps/packages/wallet/src/domain/commitments/__tests__/index.ts`. 

For the typescript that is to live in the wallet package itself, we have the option to leverage some simplifications from only having two players in the consensus game running in a ledger channel (since that is all that we will support for now). On the other hand, the new consensus app is sufficiently intuitive that we may not need to do that (for example, `vote` is never used, only `finalVote`).


packages/wallet/src/domain/two-player-consensus-game/index.ts
```diff
@@ -0,0 +1,199 @@
...
+): commitment is ConsensusCommitment {
+  return commitment.consensusCounter === 0;
+}
+export function isProposal(
+  commitment: ConsensusGameBaseCommitment,
+): commitment is ProposalCommitment {
+  return commitment.consensusCounter === 1;
+}
+
+export function nextLedgerUpdateCommitment(
```
tomclose: I'm not sure we want this method. I'd prefer to always ensure we use the methods above in the code. Could you remove this and replace the places where it's called?

packages/wallet/src/domain/two-player-consensus-game/index.ts
```diff
@@ -0,0 +1,199 @@
...
+// TODO
+// function initialize(channelDetails: ??, initialOutcome: Outcome): Proposal
+
+export function accept(commitment: ProposalCommitment): ConsensusCommitment {
+  return consensusCommitment({
+    ...commitment,
+    turnNum: commitment.turnNum + 1,
+    allocation: commitment.proposedAllocation,
+    destination: commitment.proposedDestination,
+    // TODO should proposedAllocation be conserved, or reset?
```
tomclose: Reset

tomclose: Or set to `0x0`

packages/wallet/src/domain/two-player-consensus-game/index.ts
```diff
@@ -0,0 +1,199 @@
...
+
+// TODO
+// function initialize(channelDetails: ??, initialOutcome: Outcome): Proposal
+
+export function accept(commitment: ProposalCommitment): ConsensusCommitment {
+  return consensusCommitment({
+    ...commitment,
+    turnNum: commitment.turnNum + 1,
+    allocation: commitment.proposedAllocation,
+    destination: commitment.proposedDestination,
```
tomclose: I'm experimenting with a `Balances` interface here: https://github.com/magmo/apps/blob/indirect-funding2/packages/wallet/src/domain/commitments/__tests__/index.ts#L40-L55.

I think this might be superior as it (a) seems more 'natural' and (b) is more readily extendable to multiple currencies later. What do you think about using it here for the first time in non-test code?

geoknee:
See also _https://github.com/magmo/force-move-games/pull/63_

packages/wallet/src/domain/two-player-consensus-game/index.ts
geoknee: The convention that we seem to have settled on is that the default values for these are *typed* but empty arrays.

PR #446, opened 2019-05-01
Create communication module

andrewgordstewart:
The communication module serves to define the messages that wallets can send to each other. This facilitates server development, which needs to understand.

I also propose that protocols use bespoke actions, like `concludeChannel`. This way, the wallet knows exactly how to route the action. Right now, it seems like the wallet would have to inspect a `CommitmentReceived` action, say "This commitment is a `CONCLUDE` commitment", check if a conclude process is running for the channel, and if not, create a new process running the `Conclude` protocol.

Once accepted, I would contain all referencers to `messageRelayRequested` by the communication module, and then write out the communication test scenarios.

### DIFF
--- a/packages/wallet/src/communication/index.ts
+++ b/packages/wallet/src/communication/index.ts
@@ -1,4 +1,5 @@
 import { Commitment } from '../domain';
+import { messageRelayRequested } from 'magmo-wallet-client';
 
 export enum Strategy {
   IndirectFunding = 'IndirectFunding',
@@ -31,7 +32,6 @@ export const strategyApproved = (processId: string): StrategyApproved => ({
 });
 
 // CONCLUDING
-
 export const CONCLUDE_CHANNEL = 'WALLET.CONCLUDING.CONCLUDE_CHANNEL';
 export interface ConcludeChannel extends BaseProcessAction {
   type: typeof CONCLUDE_CHANNEL;
@@ -48,3 +48,9 @@ export const concludeChannel = (
   commitment,
   signature,
 });
+
+export type Message = StrategyProposed | StrategyApproved | ConcludeChannel;
+export function sendMessage(to: string, message: Message) {
+  const { processId } = message;
+  return messageRelayRequested(to, { processId, data: message });
+}
--- a/packages/wallet/src/communication/index.ts
+++ b/packages/wallet/src/communication/index.ts
@@ -1,0 +1,8 @@
+export enum Strategy {
+  IndirectFunding = 'IndirectFunding',
+}
+
+export interface BaseProcessAction {
+  processId: string;
+  type: string;
+}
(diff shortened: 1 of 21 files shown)


packages/wallet/src/communication/index.ts
```diff
@@ -0,0 +1,56 @@
...
+  signature: string,
+): ConcludeChannel => ({
+  type: CONCLUDE_CHANNEL,
+  processId,
+  commitment,
+  signature,
+});
+
+export type Message = StrategyProposed | StrategyApproved | ConcludeChannel;
+export function sendMessage(to: string, message: Message) {
```
tomclose: Do you think it's true that this will basically always be called with one of the methods above? If so, it might be worth skipping the middleman and defining `sendConcludeChannel` etc.

packages/wallet/src/communication/index.ts
```diff
@@ -0,0 +1,56 @@
+import { Commitment } from '../domain';
+import { messageRelayRequested } from 'magmo-wallet-client';
+
+export enum Strategy {
+  IndirectFunding = 'IndirectFunding',
+}
```
tomclose: It seems like `type Strategy = 'IndirectFunding' | ... ` might be superior here. Having a type makes it easier to combine these options or add more complex options (e.g. `IndirectFunding(with_buffer: 20%)`) later. VSCode gives you all the auto-complete goodness.

packages/wallet/src/communication/index.ts
andrewgordstewart: It should only be called with those methods above, yes.

packages/wallet/src/communication/index.ts
andrewgordstewart: There are, on the other hand, some downsides to a discriminated union:
You can do this:
```
export const INDIRECT = 'IndirectFundingStrategy'
export type INDIRECT = typeof INDIRECT;
export type FundingStrategy = INDIRECT;
```
which gets verbose, for no clear benefit if you don't have complex types.

Or you can do this: 
```
export type FundingStrategy = 'IndirectFundingStrategy';
```

in which case you need to do things like
```
const strategy = 'IndirectFundingStrategy' as 'IndirectFundingStrategy';
// or:
// const strategy: FundingStrategy = 'IndirectFundingStrategy';
const waitForStrategyResponse = states.waitForStrategyResponse(props);
```
because Typescript does not know that `strategy` was supposed to be of type `FundingStrategy`, and assigns it type `string`.

The enum approach has some other benefits:
- you can iterate over the values of an enum (perhaps useful in views)
- The values of an enum are tightly tied to the enum type itself: The only way to access `Strategy.IndirectFunding` is through the `Strategy` enum, while the string `'IndirectFundingStrategy'` can be used in any context that expects a string.

I've tried switching to the simple union type of anonymous strings.

PR #453, opened 2019-05-03
Split concluding protocol

geoknee:
...into instigator and responder subprotocols. Looking for some early feedback on this, and to discuss if it's the right approach. 

So far the plan is encapsulated by the new directory structure and new readme files.


packages/wallet/src/redux/protocols/concluding/instigator/readme.md
```diff
@@ -0,0 +1,80 @@
...
+  Su((Success))
+  NT2(More Non-Terminal States) --> |FAILURE_TRIGGER| F
+  F((Failure))
+
+  style St  fill:#efdd20
+  style L fill:#efdd20
+  style C stroke:#333,stroke-width:4px
+  style Su fill:#58ef21
+  style F  fill:#f45941
+```
```
tomclose: You probably don't need this to understand the diagram. Might be useful as part of a contribution guidelines or similar though.

packages/wallet/src/redux/protocols/concluding/responder/readme.md
```diff
@@ -0,0 +1,80 @@
+# Concluding Protocol
```
tomclose: I'm guessing you haven't updated this file yet?

geoknee: I thought I had...

packages/wallet/src/redux/protocols/concluding/instigator/readme.md
geoknee: Agreed: it's also in some of the other protocol readmes. At some point we should do some tidying up across the protocols, and absorb this into a top level readme.

packages/wallet/src/redux/actions.ts
```diff
@@ -158,7 +158,8 @@ export type ProtocolAction =
   | RespondingAction
   | application.ApplicationAction
   | DefundingAction
-  | ConcludingAction;
+  | ConcludingActionI
```
tomclose: Probably worth making this `ConcludingInstigatorAction`

PR #458, opened 2019-05-04
Funding protocol fixes

andrewgordstewart:
- The `FUNDING_REQUESTED` action starts the `'Funding'` protocol, not the `'IndirectFunding'` protocol
- The actions that choose and accept a strategy are correctly modified
- The `opponentAddress` is correctly set on the funding state


packages/wallet/src/redux/protocols/funding/reducer.ts
```diff
@@ -18,7 +18,8 @@ export function initialize(
 ): ProtocolStateWithSharedData<states.FundingState> {
   // TODO: We probably want to handle this in the root reducer
   const channelState = selectors.getChannelState(sharedData, channelId);
-  const opponentAddress = channelState.participants[playerIndex];
+  const { participants } = channelState;
+  const opponentAddress = participants[playerIndex + (1 % participants.length)];
```
lalexgap: This could probably be a helper function `getOpponentAddress` that is shared between reducers.

PR #459, opened 2019-05-05
Add Indirect Funding Sub-protocol to funding protocol

lalexgap:
Simple change so the funding protocol properly initializes the indirect funding sub-protocol and delegates appropriate actions to it.

### DIFF
--- a/packages/wallet/src/redux/protocols/funding/player-a/reducer.ts
+++ b/packages/wallet/src/redux/protocols/funding/player-a/reducer.ts
@@ -80,7 +80,7 @@ function handleIndirectFundingAction(
     sharedData,
     action,
   );
-  if (!indirectFundingStates.isTerminal(newProtocolState.fundingState)) {
+  if (indirectFundingStates.isTerminal(newProtocolState.fundingState)) {
     return { protocolState: handleFundingComplete(newProtocolState), sharedData: newSharedData };
   } else {
     return { protocolState: newProtocolState, sharedData: newSharedData };
--- a/packages/wallet/src/redux/protocols/funding/player-a/reducer.ts
+++ b/packages/wallet/src/redux/protocols/funding/player-a/reducer.ts
@@ -112,8 +112,15 @@ function strategyApproved(
   if (state.type !== states.WAIT_FOR_STRATEGY_RESPONSE) {
     return { protocolState: state, sharedData };
   }
-
-  return initializeFunding(state, sharedData);
+  const channelState = selectors.getChannelState(sharedData, state.targetChannelId);
+  const { protocolState: fundingState, sharedData: newSharedData } = initializeIndirectFunding(
+    channelState,
+    sharedData,
+  );
+  return {
+    protocolState: states.waitForFunding({ ...state, fundingState }),
+    sharedData: newSharedData,
+  };
 }
 
 function strategyRejected(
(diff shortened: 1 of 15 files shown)


packages/wallet/src/redux/protocols/funding/player-a/reducer.ts
```diff
@@ -143,3 +173,38 @@ function cancelled(state: states.FundingState, sharedData: SharedData, action: a
...
+  protocolState: states.WaitForFunding,
+): states.WaitForSuccessConfirmation | states.Failure {
+  if (protocolState.fundingState.type === 'Success') {
+    return states.waitForSuccessConfirmation(protocolState);
+  } else {
+    // TODO: Indirect funding should return a proper error to pass to our failure state
+    return states.failure('Funding Failure');
+  }
+}
+function updateFundingState(
```
andrewgordstewart: It looks like there are two identical functions called `updateFundingState` introduced. Is there a reason for that?

packages/wallet/src/redux/protocols/funding/player-a/states.ts
```diff
@@ -36,7 +37,7 @@ export interface WaitForStrategyResponse extends BaseState {
 
 export interface WaitForFunding extends BaseState {
   type: typeof WAIT_FOR_FUNDING;
-  fundingState: 'funding state';
+  fundingState: IndirectFundingState;
```
andrewgordstewart: Should terminal funding states be included in the type of `fundingState`?

packages/wallet/src/redux/protocols/funding/player-b/container.tsx
```diff
@@ -36,8 +38,11 @@ class FundingContainer extends PureComponent<Props> {
           />
         );
       case states.WAIT_FOR_FUNDING:
-        // TODO: embed the funding container
-        return <div />;
+        if (isTerminal(state.fundingState)) {
```
andrewgordstewart: This check is required since the compiler doesn't know that `state.fundingState` is never terminal. Should that ever occur?

packages/wallet/src/redux/protocols/indirect-funding/state.ts
```diff
@@ -22,3 +22,7 @@ export function success(): Success {
 export function failure(): Failure {
   return { type: 'Failure' };
 }
+
+export function isTerminal(state: IndirectFundingState): state is Failure | Success {
+  return state.type === 'Failure' || state.type === 'Success';
```
andrewgordstewart: This is a case where "magic strings" can lead to an unsafe type guard. It's possible to change the type of `Failure` to `{ type: 'FAILURE' }`, which will render this type guard's implementation incorrect. The type system will not notice this, and there will be no compiler error.

lalexgap: Since state is typed as `IndirectFundingState` the compiler will complain if the comparison `state.type === 'Failure'` doesn't include a valid type. So if we change `Failure` to`FAILURE` we get a compile error right away.

packages/wallet/src/redux/protocols/funding/player-a/reducer.ts
lalexgap: One for player A and one for player B. They are slightly different in that they are transitioning to the relevant player A or B states  at the end.

packages/wallet/src/redux/protocols/funding/player-a/states.ts
lalexgap: I think you're right that we don't need to include terminal states. I've updated this. We should probably be consistent and only use non-terminal states for other sub protocols as well.

packages/wallet/src/redux/protocols/funding/player-b/container.tsx
lalexgap: Updated

packages/wallet/src/redux/protocols/funding/player-a/reducer.ts
lalexgap: This has been obsoleted regardless with the latest change.

packages/wallet/src/redux/protocols/indirect-funding/state.ts
andrewgordstewart: Right, nice! It is still possible to have an issue, where one of the other `IndirectFundingState` has type `'Failure'`. But that's just an academic example and would be a much bigger problem. (Using an enum would eliminate that possibility.)

Issue #461, opened 2019-05-06
Server and client wallets/apps need to share prebuilt artifacts


kerzhner:
The plan is to create a separate package.

PR #464, opened 2019-05-07
New package for shared data

kerzhner:
The intention of the new package is to contain data shared between the client wallet/applications, server wallet/applications, or any other wallet/application implementations. Also, including `path.js` cleanup.

Note: `splice-networks.sh` script will no longer work. The way `test` and `development` environments are set up will change soon, so ignoring that for now.

Fixes https://github.com/magmo/apps/issues/461.


kerzhner:
Thanks for the feedback @andrewgordstewart. Addressing in bf0e39bcbc7b14d8defb69699998cfce6347bf42

PR #465, opened 2019-05-07
Indirect Funding Updates

lalexgap:
This PR finishes up the player A reducer to get the player A tests passing.

I've also started working on the app to app funding. Currently I've gotten it to the point that the player A deposit gets sent to metamask.

TODO:

Currently the adjudicator watcher relies on `channelsToMonitor` to dispatch events. However there is no mechanism for a protocol reducer to update `channelsToMonitor`. `channelsToMonitor` gets initialized to an empty array which means no `FUNDING_RECEIVED_EVENT` gets dispatched by the adjudicator watcher. Which means the direct funding reducer gets stuck waiting for the `FUNDING_RECEIVED_EVENT`.

We need some way for the indirect funding reducer to update `channelsToMonitor` once we create the ledger channel, so `FUNDING_RECEIVED_EVENT`s get generated. That or we need the adjudicator watcher to always dispatch actions and rely on reducers to determine if a adjudicator action applies to one of its channels.

### DIFF
--- a/packages/wallet/src/redux/protocols/indirect-funding/player-b/reducer.ts
+++ b/packages/wallet/src/redux/protocols/indirect-funding/player-b/reducer.ts
@@ -35,6 +35,7 @@ import { directFundingStateReducer } from '../../direct-funding/reducer';
 import { isSuccess, isFailure } from '../../direct-funding/state';
 import { acceptConsensus } from '../../../../domain/two-player-consensus-game';
 import { sendCommitmentReceived } from '../../../../communication';
+import { addHex } from '../../../../utils/hex-utils';
 
 type ReturnVal = ProtocolStateWithSharedData<IndirectFundingState>;
 type IDFAction = actions.indirectFunding.Action;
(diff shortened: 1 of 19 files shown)


packages/wallet/src/redux/protocols/indirect-funding/player-b/reducer.ts
```diff
@@ -201,14 +208,8 @@ function handleWaitForLedgerUpdate(
   // are we happy that we have the ledger update?
   // if so, we need to craft our reply
 
-  const theirConsensusCommitment = fromCoreCommitment(theirCommitment);
-  if (theirConsensusCommitment.updateType !== UpdateType.Proposal) {
-    throw new Error('The received commitment was not a ledger proposal');
-  }
-  const ourConsensusCommitment = finalVote(theirConsensusCommitment);
+  const ourCommitment = acceptConsensus(theirCommitment);
   // TODO this should happen automatically in the finalVote helper function
```
geoknee:
```suggestion
```

Issue #437, opened 2019-04-30
Standalone, adjudicator watcher server process


kerzhner:
- [x] New process for monitoring funding events and challenges.
- [x] Factor out server components shared by the adjudicator watcher and the http server.
- [x] Tests for the adjudicator watcher.
- [x] CI setup for new tests that rely on ganache running.
- [x] Heroku setup for the adjudicator watcher.

PR #466, opened 2019-05-07
Server adjudicator watcher

kerzhner:
This PR addresses part of https://github.com/magmo/apps/issues/437.

- Adds `channel_id` to the `AllocatorChannel` table.
- A standalone process monitors for chain events.
- A basic unit test that spins up the process, sends an event to Ganache and waits for the event listener to fire.
- Dynamically import locally-built artifacts vs prebuilt artifacts based on the environment.

Remaining items that can be addressed in future pull requests:
- [ ] Event monitors and tests for events other than funding.
- [ ] Spin up the adjudicator watcher when deploying to Heroku.
- [x] Local setup for manual/integration testing of the server and wallet pointing to the same Ganache instance.


packages/server/src/wallet/adjudicator-watcher/__chain-test__/adjudicator-watcher.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    `http://localhost:${process.env.DEV_GANACHE_PORT}`,
+  );
+
+  it('should handle a funds received event when channel is in the database', async done => {
+    const preEventHoldings = await getHoldings();
+
+    const eventCallback = async eventType => {
+      const postEventHoldings = await getHoldings();
+
+      const eventDeposit = bigNumberify(postEventHoldings).sub(bigNumberify(preEventHoldings));
```
tomclose: I found this name a bit confusing. Maybe something like `depositedAmount` would be clearer?

packages/server/src/wallet/adjudicator-watcher/__chain-test__/utils.ts
```diff
@@ -0,0 +1,45 @@
...
+function createDepositTransaction(destination: string, DepositLevel: string) {
+  const adjudicatorInterface = getAdjudicatorInterface();
+  const data = adjudicatorInterface.functions.deposit.encode([destination]);
+  return {
+    value: DepositLevel,
+    data,
+  };
+}
+
+export async function depositContract(
```
tomclose: Maybe `depositIntoContract` or `makeDeposit`?

packages/server/src/wallet/adjudicator-watcher/__chain-test__/adjudicator-watcher.test.ts
kerzhner: Thanks for the feedback, @tomclose.

PR #479, opened 2019-05-09
Concluding Work

lalexgap:
Work on getting concluding work. 

- Updated magmo-wallet-client to remove the `processId` from messages. This allows us to treat the `concludeInstigated` as both a message action and a new process action.
- Switched from  `ConcludingInstigator`/`ConcludingResponder` to a single `Concluding` protocol. This allows communication between the concluding players to use the same `processId` (like or other protocols)
- Updated 'responder' to initialize with a `commitment` which gets stored and validated.
- Updated the tests a bunch.
- Renamed and removed some actions.

TODO:

- [x] Update Instigator to handle `COMMITMENT_RECEIVED`
- [x] Add back unhappy path tests.
- [ ] Set `FundingState` during the funding process. We rely on the funding state to figure out whether a channel is directly funded and what the ledger id is. Currently this is never set.
- [ ] Update defunding reducer to call the `indirectDefundingReducer`

### DIFF
--- a/packages/wallet/src/redux/sagas/message-listener.ts
+++ b/packages/wallet/src/redux/sagas/message-listener.ts
@@ -72,9 +72,7 @@ export function* messageListener() {
 }
 
 function handleIncomingMessage(action: incoming.ReceiveMessage) {
-  const { messagePayload } = action as incoming.ReceiveMessage;
-
-  const { data } = messagePayload;
+  const data = action.messagePayload;
 
   if ('type' in data && isRelayableAction(data)) {
     return data;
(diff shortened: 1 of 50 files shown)


packages/wallet/src/redux/sagas/message-listener.ts
```diff
@@ -72,9 +72,7 @@ export function* messageListener() {
 }
 
 function handleIncomingMessage(action: incoming.ReceiveMessage) {
-  const { messagePayload } = action as incoming.ReceiveMessage;
-
-  const { data } = messagePayload;
+  const data = action.messagePayload;
```
geoknee: Should a test have caught this?

geoknee: (this is a fix for funding, which was broken)

PR #484, opened 2019-05-13
Fix Direct Funding Race Condition

lalexgap:
If player B received `postFundSetup0` before the channel is funded the commitment was not stored anywhere.

This means that after the channel gets funded when Player B tries to send out `postFundSetup1` it fails since we're trying to transition from `preFundSetup0` to `postFundSetup1`.

To fix this we now update the channel state with the `postFundSetup0` commitment even if the channel is not funded yet.


packages/wallet/src/redux/protocols/direct-funding/reducer.ts
```diff
@@ -159,13 +159,19 @@ const commitmentReceivedReducer: DFReducer = (
       sharedData: sharedDataWithOwnCommitment,
     };
   } else {
+    const checkResult = checkAndStore(sharedData, action.signedCommitment);
```
kerzhner: Looks like `checkAndStore` is performed (and the result is inspected) for all player B code paths. We can probably factor that logic outside of the player B `if/else` (in other words, to the top of the player B codepath).

PR #487, opened 2019-05-14
Move ConcludeInstigated to communication module

andrewgordstewart:
Actions that are communicated should be defined in the communication module.

### DIFF
--- a/packages/wallet/src/communication/actions.ts
+++ b/packages/wallet/src/communication/actions.ts
@@ -1,7 +1,7 @@
 import { SignedCommitment } from '../domain';
 import { WalletAction } from '../redux/actions';
 import { FundingStrategy } from './index';
-import { CONCLUDE_INSTIGATED, ConcludeInstigated } from '../redux/protocols/actions';
+import { WalletProtocol } from '../redux/types';
 
 export interface BaseProcessAction {
   processId: string;
@@ -31,6 +31,15 @@ export const strategyApproved = (processId: string): StrategyApproved => ({
   type: STRATEGY_APPROVED,
   processId,
 });
+
+export const CONCLUDE_INSTIGATED = 'WALLET.NEW_PROCESS.CONCLUDE_INSTIGATED';
+export const concludeInstigated = (signedCommitment: SignedCommitment, channelId: string) => ({
+  type: CONCLUDE_INSTIGATED as typeof CONCLUDE_INSTIGATED,
+  signedCommitment,
+  protocol: WalletProtocol.Concluding,
+  channelId,
+});
+export type ConcludeInstigated = ReturnType<typeof concludeInstigated>;
 
 // COMMON
 export const COMMITMENT_RECEIVED = 'WALLET.COMMON.COMMITMENT_RECEIVED';
--- a/packages/wallet/src/communication/actions.ts
+++ b/packages/wallet/src/communication/actions.ts
@@ -37,7 +37,7 @@ export const concludeInstigated = (signedCommitment: SignedCommitment, channelId
   type: CONCLUDE_INSTIGATED as typeof CONCLUDE_INSTIGATED,
   signedCommitment,
   protocol: WalletProtocol.Concluding,
-  channelId,
+  processId: `${WalletProtocol.Concluding}-${channelId}`,
 });
 export type ConcludeInstigated = ReturnType<typeof concludeInstigated>;
 
--- a/packages/wallet/src/communication/actions.ts
+++ b/packages/wallet/src/communication/actions.ts
@@ -37,7 +37,7 @@ export const concludeInstigated = (signedCommitment: SignedCommitment, channelId
   type: CONCLUDE_INSTIGATED as typeof CONCLUDE_INSTIGATED,
   signedCommitment,
   protocol: WalletProtocol.Concluding,
-  processId: `${WalletProtocol.Concluding}-${channelId}`,
+  channelId,
 });
 export type ConcludeInstigated = ReturnType<typeof concludeInstigated>;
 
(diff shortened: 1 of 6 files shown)


geoknee:
This function https://github.com/magmo/apps/blob/52dcdc96d05d44c6c9778dd549d10989cef126e6/packages/wallet/src/redux/reducer.ts#L133-L137 is relied on in the top level reducer *and* in the test of that reducer. This leads to the undesirable property that not setting a `channelId` on your action does not cause any error nor any test to fail, despite it actually breaking some functionality of the wallet. The problem is that the responder wallet gets the default `processId` when we want it to be `Concluding-0x4uygu...`.

I propose that we remove that default case and/or toughen up the test of the top level reducer. 

As for whether we should be sending a `processId` or a `channelId` on the `ConcludeInstigated` action, to me it makes slightly more sense for it to be the latter. This is because the content of that message is, I as the instigator am concluding that particular channel.  In the end it won't make much difference if we insist on a consistent way to infer the `processId` from the `channelId` and the protocol type: it is just a question of whether the instigator or the responder does that computation. But in some different model of how the wallet works, instigator and responder could have different `processId`s governing concluding of the same channel with `channelId`; in that model we would want the responder to compute the `processId` in their own preferred way

packages/wallet/src/communication/actions.ts
```diff
@@ -32,6 +32,15 @@ export const strategyApproved = (processId: string): StrategyApproved => ({
   processId,
 });
 
+export const CONCLUDE_INSTIGATED = 'WALLET.NEW_PROCESS.CONCLUDE_INSTIGATED';
+export const concludeInstigated = (signedCommitment: SignedCommitment, channelId: string) => ({
+  type: CONCLUDE_INSTIGATED as typeof CONCLUDE_INSTIGATED,
+  signedCommitment,
+  protocol: WalletProtocol.Concluding,
+  processId: `${WalletProtocol.Concluding}-${channelId}`,
```
andrewgordstewart: @geoknee Adding a comment here so we can have a proper discussion

> This function
> 
> https://github.com/magmo/apps/blob/52dcdc96d05d44c6c9778dd549d10989cef126e6/packages/wallet/src/redux/reducer.ts#L133-L137
> 
> is relied on in the top level reducer _and_ in the test of that reducer. This leads to the undesirable property that not setting a `channelId` on your action does not cause any error nor any test to fail, despite it actually breaking some functionality of the wallet. The problem is that the responder wallet gets the default `processId` when we want it to be `Concluding-0x4uygu...`.
> I propose that we remove that default case and/or toughen up the test of the top level reducer.
> 
> As for whether we should be sending a `processId` or a `channelId` on the `ConcludeInstigated` action, to me it makes slightly more sense for it to be the latter. This is because the content of that message is, I as the instigator am concluding that particular channel. In the end it won't make much difference if we insist on a consistent way to infer the `processId` from the `channelId` and the protocol type: it is just a question of whether the instigator or the responder does that computation. But in some different model of how the wallet works, instigator and responder could have different `processId`s governing concluding of the same channel with `channelId`; in that model we would want the responder to compute the `processId` in their own preferred way

andrewgordstewart: So, there are a few things that should come out of this:
1. Ideally, this error should be caught by either a type error, through some convention that we use, or through a test.
2. The following is wrong.
>we would want the responder to compute the processId in their own preferred way

For my wallet to communicate with your wallet, we need to compute the `processId` in the _same_ way, not in our _preferred_ way. Therefore, the `getProcessId` function is a part of the "magmo wallet protocol" -- ie. proper communication relies on it -- and thus, it should be exported from the "`wallet` library" (`./index.ts`)
 Ideally, this code can live in its own package, but right now the web app and the core code is very coupled.
3. There should be some way for the recipient of `ConcludeInstigated` to know that it needs to create a new process fro this action. That could simply be "check if the type is `'CONCLUDE_INSTIGATED'`", but ideally this would also be defined as common code.

I don't think either of these things have to be addressed right now, except for reverting the relevant commit.

geoknee: 1. Agreed
2. In our current approach, yes the `processId` should match in all participant's wallets and I would not suggest changing that. The point I was trying to make is it is *more sure* that the `channelId` should be on the message, than it is the `processId`. Imagine for instance that our wallet is communicating with some third party wallet that also follows Nitro protocol (but not necessarily "magmo wallet protocol"). Of course that is way out of scope, but it was part of my reasoning for preferring `channelId`. 
3.  It is a member of the `NewProcessAction` type and can be discriminated as such.

As quick fix for 1 would be to replace the default `return APPLICATION_PROCESS_ID;` with an error throw (and followup on the tests that will now fail). What do you think?

andrewgordstewart: > 2. Imagine for instance that our wallet is communicating with some third party wallet that also follows Nitro protocol (but not necessarily "magmo wallet protocol").

But, for two parties to communicate, they need to agree on a means of communication. We should be specifying that means of communication!

> 3. It is a member of the `NewProcessAction` type and can be discriminated as such.

`NewProcessAction` is just a typescript type, and does not exist at runtime. More so, neither it nor the type guard implemented in the wallet package are appropriate in the context of communication, as it is internal to how the client wallet works.

Given that, it might actually be the responsibility of the wallet to define its own such type/type guard.

> As quick fix for 1 would be to replace the default `return APPLICATION_PROCESS_ID;` with an error throw (and followup on the tests that will now fail). What do you think?

Yes, this seems fine, and probably just good engineering, though you should still return `APPLICATION_PROCESS_ID` when appropriate

geoknee:
Thanks @andrewgordstewart ! Seems good to separate actions that are triggered by messages received from other parties.

Summary: these actions now live in a more appropriate place, have their own type guard and have their own `getProcessId` helper. Moreover, the omission of `channelId` will cause a runtime error to be thrown by the recipient. 👍

PR #490, opened 2019-05-15
Environment variable revamp

kerzhner:
This PR builds from https://github.com/magmo/devtools/pull/25/files. Once that PR is merged we should create a new `magmo-devtools` tag.

Changes for this PR:
- No longer set `NODE_ENV` in source code. Doing so breaks the [hierarchy of environment variable loading](https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use).
- Use `dotEnv` from `magmo-devtools`.


packages/server/.env.test
```diff
@@ -0,0 +1 @@
+SERVER_NETWORK_ID=123456789
```
andrewgordstewart: We should probably rename this at some point, since `SERVER` is pretty ambiguous.

packages/server/.vscode/launch.json
```diff
@@ -18,7 +18,10 @@
       "cwd": "${workspaceRoot}",
       "protocol": "inspector",
       "console": "integratedTerminal",
-      "internalConsoleOptions": "neverOpen"
+      "internalConsoleOptions": "neverOpen",
+      "env": {
+        "NODE_ENV": "test"
```
andrewgordstewart: :+1: nice

packages/server/.env.test
kerzhner: Agreed that `SERVER` is ambiguous. `SERVER_NETWORK_ID` can be Ganache, Ropsten, Mainnet, etc. So maybe `CHAIN_NETWORK_ID`?

PR #494, opened 2019-05-16
Upgrade storybook to fix storybook error

lalexgap:
Upgrade to latest version 4 of storybook to fix the error we were running into. There's a version 5 but that doesn't seem to work with our current storybook config, so I just upgraded to the latest version 4.


lalexgap:
> @lalexgap do you think we should also upgrade Storybook in the `rps` package?

Good point, I'll upgrade `rps` as well.

PR #499, opened 2019-05-17
Update server to communicate via protocol actions.

andrewgordstewart:
(Apologies for the rather large, unfocused PR)

This PR sets up the server to progress an rps channel to the application stage (and lets you keep playing). Concluding was working for a bit, but is currently broken for me (from rebasing?).

Here are the highlights
- Adds a `v2/channels` route
- Adds functionality to route some of the actions defined in the `magmo-wallet/communication` module
- Contains some small changes to migrations
- Exports some things from the wallet package that are used by the server
- Currently works 
- signature checking is temporarily disabled

It probably doesn't work without some local environment configuration, based on changes made to devtools/apps.


packages/server/src/app/routes/v2/channels.ts
```diff
@@ -0,0 +1,56 @@
...
+      return await handleNewProcessAction(ctx);
+    } else if (await isProtocolAction(action)) {
+      return await handleOngoingProcessAction(ctx);
+    }
+  }
+});
+
+export const channelRoutes = router.routes();
+
+async function isNewProcessAction(action: RelayableAction): Promise<boolean> {
```
kerzhner: Does this need to be asynchronous?

packages/server/src/app/handlers/handle-new-process-action.ts
```diff
@@ -0,0 +1,42 @@
...
+  const action = ctx.request.body;
+  if (!communication.isRelayableAction(action)) {
+    throw new Error(`Action ${action.type} is not valid`);
+  }
+
+  switch (action.type) {
+    case 'WALLET.COMMON.COMMITMENT_RECEIVED':
+    case 'WALLET.FUNDING.STRATEGY_PROPOSED':
+    case 'WALLET.FUNDING.STRATEGY_APPROVED':
+      return ctx;
```
kerzhner: Is the path for these actions yet to be implemented? Or is this the implemented codepath?

packages/server/src/app/handlers/handle-ongoing-process-action.ts
```diff
@@ -0,0 +1,95 @@
...
+    case 'WALLET.FUNDING.STRATEGY_PROPOSED':
+      return handleStrategyProposed(ctx, action);
+  }
+}
+
+async function handleStrategyProposed(ctx, action: StrategyProposed) {
+  const { processId } = action;
+  const process = await getProcess(processId);
+  if (!process) {
+    throw errors.processMissing(processId);
```
kerzhner: Do you know if the error is returned as part of the response to the client?

packages/server/src/app/handlers/handle-ongoing-process-action.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    if (process.env.NODE_ENV !== 'test') {
+      // TODO: Figure out how to test this.
+      const expectedHeld =
+        theirCommitment.allocation[1 - theirCommitment.destination.indexOf(HUB_ADDRESS)];
+      const funding = theirCommitment.allocation[theirCommitment.destination.indexOf(HUB_ADDRESS)];
+
+      setTimeout(async () => {
+        // For the moment, we delay the deposit to give the user a chance to deposit.
+        await Blockchain.fund(channelID(theirCommitment.channel), expectedHeld, funding);
```
kerzhner: Is the server kicking off funding on every received ledger channel commitment?

packages/server/src/app/handlers/handle-new-process-action.ts
andrewgordstewart: Yes, it's still being fleshed out.

packages/server/src/app/handlers/handle-ongoing-process-action.ts
andrewgordstewart: Not in this case. I haven't quite figured out how to manage error handling, and wasn't concerned with it at the moment

PR #507, opened 2019-05-22
Server Deposit Manager

kerzhner:
The goal of this PR is to remove the time delayed server deposit. This PR:
- Minimizes functionality of the adjudicator watcher. The adjudicator watcher is now only responsible for monitoring for blockchain events and forwarding the events to the application server.
- Adds deposit manager that acts as needed when a blockchain deposit arrives.
- Blockchain functionality is mocked out so that the deposit manager unit tests are independent of ganache state.

As for testing:
- The deposit manager is poorly tested. I am having trouble figuring out how to spy on es6 class static functions. Specifically, it would be nice to spy on `Blockhain.fund`. 
- The deposit manager has not been end-to-end, manually tested.


andrewgordstewart:
I merged https://github.com/magmo/apps/pull/514, which was rebased onto master.

PR #508, opened 2019-05-22
Minor clean up changes

lalexgap:


packages/wallet/src/__stories__/index.tsx
```diff
@@ -31,7 +33,17 @@ export const protocolStateRender = (protocolState: ProtocolState) => {
 
   return walletStateRender(walletState);
 };
-
+function nestProtocolState(protocolState: ProtocolState): ProtocolState {
```
lalexgap: @geoknee This is my first stab at trying to get stories for sub-protocols to render underneath their parent protocol. I think we could expand this to handle the other sub-protocols but it's going to get a bit messy constructing so many different parent protocol states.

geoknee: Thanks @lalexgap . I finished this off by: 

1. Making a diagram showing the hierarchy of protocols
2. Writing some 'Nester' functions that live on some of the parent protocol's `states.ts`
3. Filling out your `nestProtocolState` function with the shortest chain of nesters from any particular protocolState to the top level protocol state

This works neatly in a lot of cases because the nestED state often has all of the information needed to construct the nestER state. In a couple of cases I had to put some inconsequential dummy values in. Aside from that problem the nester functions could be used outside of testing, too.

geoknee: Screenshot 2019-05-23 at 13 36 39

PR #509, opened 2019-05-24
Re-Use Existing Ledger Channel

### LINKED ISSUE #504: Re-use ledger channels for multiple serial games
- Do not automatically defund the ledger channel when the application concludes
- Allow an existing ledger channel to be used when an application requests funding.

lalexgap:
This PR allows for existing ledger channels to be reused in a simple manner.

The funding protocol now checks for an existing ledger channel and starts the existing-channel funding protocol. If necessary the ledger-top-up protocol will run to add funds to the ledger channel so it can fund the app channel.

The concluding protocol has also been updated to allow users to keep the channel open after completing a game so it can be re-used. Both players must decide to keep the channel open.

Direct funding has also been updated to have an option whether or not to exchange post fund setups.

Some things to  be tackled in separate PRs:
- Add any missing screens / Change wording to be more appropriate
- Move Existing Channel Funding to be started from Indirect-Funding instead of Funding

Fixes #504 and #537

### DIFF
--- a/packages/rps/src/redux/message-service/saga.ts
+++ b/packages/rps/src/redux/message-service/saga.ts
@@ -66,7 +66,11 @@ export function* sendWalletMessageSaga() {
         console.error(err);
       }
     } else {
-      yield call(reduxSagaFirebase.database.create, `/messages/${to.toLowerCase()}`, messageToSend);
+      yield call(
+        reduxSagaFirebase.database.create,
+        `/messages/${to.toLowerCase()}`,
+        sanitizeMessageForFirebase(messageToSend),
+      );
     }
   }
 }
@@ -127,7 +131,7 @@ export function* sendMessagesSaga() {
         yield call(
           reduxSagaFirebase.database.create,
           `/messages/${opponentAddress.toLowerCase()}`,
-          toSend,
+          sanitizeMessageForFirebase(toSend),
         );
         yield put(gameActions.messageSent());
       }
@@ -388,4 +392,8 @@ function* signMessage(commitment: Commitment) {
       throw new Error(err.error);
     }
   }
+}
+
+function sanitizeMessageForFirebase(message) {
+  return JSON.parse(JSON.stringify(message));
 }
--- a/packages/rps/src/redux/message-service/saga.ts
+++ b/packages/rps/src/redux/message-service/saga.ts
@@ -274,6 +274,7 @@ function* handleWalletMessage(walletMessage: WalletRequest, state: gameStates.Pl
         yield put(gameActions.messageSent());
         yield put(gameActions.exitToLobby());
       } else {
+        yield put(gameActions.messageSent());
         if (concludeResponse.reason !== 'UserDeclined') {
           throw new Error(concludeResponse.error);
         }
--- a/packages/wallet/src/redux/protocols/concluding/responder/readme.md
+++ b/packages/wallet/src/redux/protocols/concluding/responder/readme.md
@@ -29,10 +29,15 @@ linkStyle default interpolate basis
   MT  --> |No| AF(ResponderAcknowledgeFailure)
   CC  --> |WALLET.CONCLUDING.RESPONDER.CONCLUDE.APPROVED| DD(ResponderDecideDefund)
   DD --> |WALLET.CONCLUDING.RESPONDER.DEFUND.CHOSEN| D(ResponderWaitForDefund)
+  DD --> |WALLET.CONCLUDING.RESPONDER.KEEP_OPEN.CHOSEN| WO(WaitForOpponentDecision)
   DD --> |COMMITMENT_RECEIVED| D(ResponderWaitForDefund)
-  D   --> |defunding protocol succeeded| AS(ResponderAcknowledgeSuccess)
+  WO --> |COMMITMENT_RECEIVED| D(ResponderWaitForDefund)
+  WO -->|WALLET.CONCLUDING.KEEP_OPEN_SELECTED|CU(WaitForLedgerUpdate)
+  CU -->|COMMITMENT_RECEIVED|CU
+  CU --> |consensus update protocol succeeded|AS
+  D  --> |defunding protocol succeeded| AS(ResponderAcknowledgeSuccess)
   AS -->  |WALLET.CONCLUDING.RESPONDER.ACKNOWLEDGED| SS((Success))
-  D   --> |defunding protocol failed| AF(ResponderAcknowledgeFailure)
+  D  --> |defunding protocol failed| AF(ResponderAcknowledgeFailure)
   classDef logic fill:#efdd20;
   classDef Success fill:#58ef21;
   classDef Failure fill:#f45941;
@@ -40,7 +45,7 @@ linkStyle default interpolate basis
   class S,E,MT logic;
   class SS Success;
   class F Failure;
-  class D WaitForChildProtocol;
+  class D,CU WaitForChildProtocol;
 ```
 
 ## Scenarios
--- a/packages/wallet/src/redux/protocols/concluding/responder/reducer.ts
+++ b/packages/wallet/src/redux/protocols/concluding/responder/reducer.ts
@@ -32,7 +32,7 @@ import {
   sendOpponentConcluded,
 } from '../../reducer-helpers';
 import { ProtocolAction } from '../../../../redux/actions';
-import { isConcludingAction } from './actions';
+import { isConcludingResponderAction } from './actions';
 import { getChannelId, SignedCommitment } from '../../../../domain';
 import { failure, success } from '../states';
 import { ProtocolStateWithSharedData } from '../..';
@@ -52,7 +52,7 @@ export function responderConcludingReducer(
     return handleDefundingAction(protocolState, sharedData, action);
   } // COMMITMENT_RECEIVED is a defunding action
 
-  if (!isConcludingAction(action)) {
+  if (!isConcludingResponderAction(action)) {
     return { protocolState, sharedData };
   }
 
--- a/packages/wallet/src/redux/protocols/concluding/responder/reducer.ts
+++ b/packages/wallet/src/redux/protocols/concluding/responder/reducer.ts
@@ -39,7 +39,12 @@ import { ProtocolStateWithSharedData } from '../..';
 import { waitForLedgerUpdate } from '../../indirect-defunding/states';
 import { waitForLedgerDefunding } from '../../defunding/states';
 import { indirectDefundingReducer } from '../../indirect-defunding/reducer';
-
+import * as helpers from '../../reducer-helpers';
+import {
+  initialize as consensusUpdateInitialize,
+  consensusUpdateReducer,
+} from '../../consensus-update/reducer';
+import { ConsensusUpdateAction, isConsensusUpdateAction } from '../../consensus-update/actions';
 export type ReturnVal = ProtocolStateWithSharedData<states.ResponderConcludingState>;
 export type Storage = SharedData;
 
@@ -48,6 +53,12 @@ export function responderConcludingReducer(
   sharedData: SharedData,
   action: ProtocolAction,
 ): ReturnVal {
+  if (
+    protocolState.type === 'ConcludingResponder.WaitForLedgerUpdate' &&
+    isConsensusUpdateAction(action)
+  ) {
+    return handleLedgerUpdateAction(protocolState, sharedData, action);
+  }
   if (isDefundingAction(action)) {
     return handleDefundingAction(protocolState, sharedData, action);
   } // COMMITMENT_RECEIVED is a defunding action
@@ -104,6 +115,42 @@ export function initialize(
     return {
       protocolState: acknowledgeFailure({ channelId, processId, reason: 'NotYourTurn' }),
       sharedData: showWallet(sharedData),
+    };
+  }
+}
+
+function handleLedgerUpdateAction(
+  protocolState: NonTerminalCState,
+  sharedData: Storage,
+  action: ConsensusUpdateAction,
+): ReturnVal {
+  if (protocolState.type !== 'ConcludingResponder.WaitForLedgerUpdate') {
+    return { protocolState, sharedData };
+  }
+  const {
+    protocolState: updatedConsensusUpdateState,
+    sharedData: newSharedData,
+  } = consensusUpdateReducer(protocolState.consensusUpdateState, sharedData, action);
+  if (updatedConsensusUpdateState.type === 'ConsensusUpdate.Success') {
+    return {
+      protocolState: states.acknowledgeSuccess(protocolState),
+      sharedData: newSharedData,
+    };
+  } else if (updatedConsensusUpdateState.type === 'ConsensusUpdate.Failure') {
+    return {
+      protocolState: states.acknowledgeFailure({
+        ...protocolState,
+        reason: 'LedgerUpdateFailed',
+      }),
+      sharedData: newSharedData,
+    };
+  } else {
+    return {
+      protocolState: states.waitForLedgerUpdate({
+        ...protocolState,
+        consensusUpdateState: updatedConsensusUpdateState,
+      }),
+      sharedData: newSharedData,
     };
   }
 }
@@ -195,9 +242,26 @@ function keepOpenChosen(protocolState: NonTerminalCState, sharedData: Storage): 
   if (protocolState.type !== 'ConcludingResponder.DecideDefund') {
     return { protocolState, sharedData };
   }
-  return {
-    protocolState: acknowledgeSuccess(protocolState),
+
+  const ledgerId = helpers.getFundingChannelId(protocolState.channelId, sharedData);
+  const appChannel = getChannel(sharedData, protocolState.channelId);
+  if (!appChannel) {
+    throw new Error(`Could not find channel ${protocolState.channelId}`);
+  }
+  const latestCommitment = appChannel.lastCommitment.commitment;
+  const {
+    protocolState: consensusUpdateState,
+    sharedData: newSharedData,
+  } = consensusUpdateInitialize(
+    protocolState.processId,
+    ledgerId,
+    latestCommitment.allocation,
+    latestCommitment.destination,
     sharedData,
+  );
+  return {
+    protocolState: states.waitForLedgerUpdate({ ...protocolState, consensusUpdateState }),
+    sharedData: newSharedData,
   };
 }
 
--- a/packages/wallet/src/redux/protocols/concluding/responder/reducer.ts
+++ b/packages/wallet/src/redux/protocols/concluding/responder/reducer.ts
@@ -230,7 +230,7 @@ function concludeApproved(protocolState: NonTerminalCState, sharedData: Storage)
       protocolState.channelId,
     );
     return {
-      protocolState: decideDefund({ ...protocolState, opponentSelectedKeepLedgerChannel: false }),
+      protocolState: decideDefund({ ...protocolState, opponentHasSelected: false }),
       sharedData: sharedDataWithOwnCommitment,
     };
   } else {
@@ -252,7 +252,7 @@ function keepOpenChosen(protocolState: NonTerminalCState, sharedData: Storage): 
     sendKeepLedgerChannelApproved(theirAddress(appChannel), protocolState.processId),
   );
 
-  if (protocolState.opponentSelectedKeepLedgerChannel) {
+  if (protocolState.opponentHasSelected) {
     const latestCommitment = getLastCommitment(appChannel);
     const {
       protocolState: consensusUpdateState,
--- a/packages/wallet/src/redux/protocols/concluding/responder/reducer.ts
+++ b/packages/wallet/src/redux/protocols/concluding/responder/reducer.ts
@@ -252,7 +252,7 @@ function keepOpenChosen(protocolState: NonTerminalCState, sharedData: Storage): 
   );
 
   if (protocolState.opponentSelectedKeepLedgerChannel) {
-    const latestCommitment = appChannel.lastCommitment.commitment;
+    const latestCommitment = getLastCommitment(appChannel);
     const {
       protocolState: consensusUpdateState,
       sharedData: newSharedData,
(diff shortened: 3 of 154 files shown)


packages/wallet/src/redux/protocols/direct-funding/reducer.ts
```diff
@@ -127,6 +137,12 @@ const fundingReceiveEventReducer: DFReducer = (
 ): ProtocolStateWithSharedData<states.DirectFundingState> => {
   // TODO[Channel state side effect]: update funding level for the channel.
 
+  if (!protocolState.exchangePostFundSetups) {
+    return {
+      protocolState: states.fundingSuccess(protocolState),
```
andrewgordstewart: This seems wrong to me, though perhaps I don't understand what `exchangePostFundSetups` means.

packages/wallet/src/redux/protocols/direct-funding/actions.ts
```diff
@@ -1,17 +1,20 @@
 import * as actions from '../../actions';
 import { ActionConstructor } from '../../utils';
+import { TwoPartyPlayerIndex } from '../../types';
```
andrewgordstewart: Why does the direct-funding protocol need to change for this PR?

packages/wallet/src/redux/protocols/existing-channel-funding/readme.md
```diff
@@ -0,0 +1,74 @@
...
+
+- Check if a ledger channel has enough funds to fund the requested channel.
+  - If not it should initiate a ledger top-up protocol.
+- Craft ledger updates to fund the channel. (Perhaps we should extract this into a protocol that can be shared by indirect-funding and existing channel funding.)
+
+Currently we assume a Ledger Top-up protocol handles both the cases where a current player needs to top-up funds and the case where their opponent needs to top-up.
+
+# State Machine
+
+### Player A State Machine
```
andrewgordstewart: I think it's sad that this will almost immediately need to be refactored to account for two-or-three-party channels.

packages/wallet/src/redux/protocols/existing-channel-funding/reducer.ts
```diff
@@ -0,0 +1,349 @@
...
+import { proposeNewConsensus, acceptConsensus } from '../../../domain/two-player-consensus-game';
+import { theirAddress, getLastCommitment } from '../../channel-store';
+import { Commitment, nextSetupCommitment } from '../../../domain';
+import { bigNumberify } from 'ethers/utils';
+import { sendCommitmentReceived } from '../../../communication';
+import { CommitmentType } from 'fmg-core';
+import { initialize as initializeLedgerTopUp, ledgerTopUpReducer } from '../ledger-top-up/reducer';
+import { isLedgerTopUpAction } from '../ledger-top-up/actions';
+export const EXISTING_CHANNEL_FUNDING_PROTOCOL_LOCATOR = 'ExistingChannelFunding';
+export const initialize = (
```
andrewgordstewart: I guess this protocol doesn't handle postFundSetup arriving before the ledger update, which probably works fine against other wallets.

packages/wallet/src/redux/protocols/existing-channel-funding/reducer.ts
```diff
@@ -0,0 +1,349 @@
...
+  proposedDestination.push(channelId);
+
+  return { proposedAllocation, proposedDestination };
+}
+
+function ledgerChannelNeedsTopUp(latestCommitment: Commitment, proposedAmount: string) {
+  if (latestCommitment.commitmentType !== CommitmentType.App) {
+    throw new Error('Ledger channel is already closed.');
+  }
+  const numParticipants = latestCommitment.channel.participants.length;
```
andrewgordstewart: I wonder if we should extract this into a helper function.

packages/wallet/src/redux/protocols/existing-channel-funding/reducer.ts
```diff
@@ -0,0 +1,349 @@
...
+  const amountRequiredFromEachParticipant = bigNumberify(proposedAmount).div(numParticipants);
```
andrewgordstewart: I don't understand this logic -- what if I already have enough funds in the channel but you don't? Then this logic says that I have to foot half of your bill?

packages/wallet/src/redux/protocols/existing-channel-funding/reducer.ts
```diff
@@ -0,0 +1,349 @@
...
+}
+
+function craftAndSendAppPostFundCommitment(
+  sharedData: SharedData,
+  appChannelId: string,
+  processId: string,
+): SharedData | 'CouldNotSign' | 'NotASetupCommitment' {
+  let newSharedData = { ...sharedData };
+  const appChannel = getChannel(sharedData, appChannelId);
+  if (!appChannel) {
```
andrewgordstewart: ✂️

packages/wallet/src/redux/protocols/existing-channel-funding/reducer.ts
```diff
@@ -0,0 +1,349 @@
...
+    throw new Error(`Could not find application channel ${appChannelId}`);
+  }
+
+  const theirAppCommitment = getLastCommitment(appChannel);
+
+  const ourAppCommitment = nextSetupCommitment(theirAppCommitment);
+  if (ourAppCommitment === 'NotASetupCommitment') {
```
andrewgordstewart: Perhaps `nextSetupCommitment` should throw instead of return a string, and always return a `Commitment`

packages/wallet/src/redux/protocols/existing-channel-funding/states.ts
```diff
@@ -0,0 +1,85 @@
...
+  type: 'ExistingChannelFunding.WaitForLedgerUpdate';
+  processId: string;
+  channelId: string;
+  ledgerId: string;
+  proposedAmount: string;
+}
+
+export interface Failure {
+  type: 'ExistingChannelFunding.Failure';
+  reason: string;
```
andrewgordstewart: Can we add type safety to the reason?

packages/wallet/src/redux/protocols/funding/player-a/reducer.ts
```diff
@@ -16,19 +16,34 @@ import {
 import * as indirectFundingStates from '../../indirect-funding/states';
 import * as selectors from '../../../selectors';
 import { Properties } from '../../../utils';
+import {
+  initialize as initializeExistingChannelFunding,
+  existingChannelFundingReducer,
+} from '../../existing-channel-funding/reducer';
```
andrewgordstewart: Can these be imported from the index?

packages/wallet/src/redux/protocols/funding/player-a/states.ts
```diff
@@ -2,13 +2,15 @@ import { ProtocolState } from '../..';
...
 import { StateConstructor } from '../../../utils';
+import { ExistingChannelFundingState } from '../../existing-channel-funding/states';
 
 // -------
 // States
 // -------
 interface BaseState {
   processId: string;
   opponentAddress: string;
+  ourAddress: string;
```
andrewgordstewart: perhaps this should be `ourAppAddress`?

packages/wallet/src/redux/protocols/funding/player-b/reducer.ts
```diff
@@ -16,19 +16,32 @@ import {
 import * as indirectFundingStates from '../../indirect-funding/states';
 import * as selectors from '../../../selectors';
 import { Properties } from '../../../utils';
-
+import {
+  isExistingChannelFundingAction,
+  ExistingChannelFundingAction,
+} from '../../existing-channel-funding/actions';
```
andrewgordstewart: index?

packages/wallet/src/redux/protocols/funding/player-b/reducer.ts
```diff
@@ -115,19 +179,58 @@ function strategyApproved(
...
+    const existingLedgerChannel = selectors.getExistingLedgerChannelForParticipants(
+      sharedData,
+      state.ourAddress,
+      state.opponentAddress,
+    );
+    if (
+      !existingLedgerChannel ||
+      getLastCommitment(existingLedgerChannel).commitmentType !== CommitmentType.App
+    ) {
+      throw new Error(
```
andrewgordstewart: Maybe in the future we want to handle this state more gracefully, right?

packages/wallet/src/redux/protocols/ledger-top-up/reducer.ts
```diff
@@ -0,0 +1,288 @@
...
+    proposedAllocation,
+    proposedDestination,
+  });
+
+  if (helpers.isFirstPlayer(ledgerId, sharedData)) {
+    const newAllocation = [
+      ...lastCommitment.allocation,
+      ...calculateTopUpAllocation(proposedAllocation, lastCommitment.allocation),
+    ];
+    const newDestination = [...lastCommitment.destination, ...proposedDestination];
```
andrewgordstewart: I don't think the rules enforce this, but I don't think multiple entries of the same participant is allowed as per the force-move protocol.

packages/wallet/src/redux/protocols/ledger-top-up/reducer.ts
```diff
@@ -0,0 +1,288 @@
...
+
+  if (!isTerminal(newDirectFundingState)) {
+    return { protocolState: newProtocolState, sharedData };
+  }
+  if (isFailure(newDirectFundingState)) {
+    return { protocolState: states.failure({ reason: 'Direct Funding Failure' }), sharedData };
+  }
+  if (isSuccess(newDirectFundingState)) {
+    const channel = getChannel(sharedData, newProtocolState.ledgerId);
+    if (!channel) {
```
andrewgordstewart: ✂️

packages/wallet/src/redux/protocols/ledger-top-up/reducer.ts
```diff
@@ -0,0 +1,288 @@
...
+    );
+    sharedData = queueMessage(sharedData, messageRelay);
+  }
+
+  // Create direct funding state
+  const lastCommitment = getLatestCommitment(sharedData, protocolState.ledgerId);
+  const directFundingAction = directFundingRequested({
+    processId: protocolState.processId,
+    channelId: protocolState.ledgerId,
+    safeToDepositLevel: isFirstPlayer ? '0x0' : lastCommitment.allocation[2],
```
andrewgordstewart: I think this kind of logic is pretty brittle, since the allocation/destination doesn't have to be in the same order as the participants.

packages/wallet/src/redux/protocols/ledger-top-up/reducer.ts
```diff
@@ -0,0 +1,288 @@
...
+};
+
+function calculateTotalTopUp(allocation: string[]): string {
+  return bigNumberify(allocation[2])
+    .add(allocation[3])
+    .toHexString();
+}
+function getOpponentAddress(sharedData: SharedData, channelId: string) {
+  const channel = getChannel(sharedData, channelId);
+  if (!channel) {
```
andrewgordstewart: ✂️

packages/wallet/src/redux/protocols/ledger-top-up/reducer.ts
```diff
@@ -0,0 +1,288 @@
...
+    throw new Error(`Could not find channel for id ${channelId}`);
+  }
+
+  return theirAddress(channel);
+}
+function getLatestCommitment(sharedData: SharedData, channelId: string) {
+  const channel = getChannel(sharedData, channelId);
+  if (!channel) {
```
andrewgordstewart: ✂️

packages/wallet/src/redux/protocols/ledger-top-up/reducer.ts
```diff
@@ -0,0 +1,288 @@
...
+}
+function getOpponentAddress(sharedData: SharedData, channelId: string) {
+  const channel = getChannel(sharedData, channelId);
+  if (!channel) {
+    throw new Error(`Could not find channel for id ${channelId}`);
+  }
+
+  return theirAddress(channel);
+}
+function getLatestCommitment(sharedData: SharedData, channelId: string) {
```
andrewgordstewart: This function doesn't seem to belong in the ledger-top-up reducer.

packages/wallet/src/redux/protocols/reducer-helpers.ts
```diff
@@ -178,3 +178,9 @@ export function getOpponentAddress(channelState: ChannelState, playerIndex: TwoP
   const opponentAddress = participants[(playerIndex + 1) % participants.length];
   return opponentAddress;
 }
+
+export function getOurAddress(channelState: ChannelState, playerIndex: TwoPartyPlayerIndex) {
+  const { participants } = channelState;
```
andrewgordstewart: What about just returning `channelState.participants[playerIndex]`?

packages/wallet/src/redux/protocols/ledger-top-up/reducer.ts
```diff
@@ -0,0 +1,288 @@
...
+  for (let i = 0; i < currentAllocation.length; i++) {
+    const currentFunds = bigNumberify(currentAllocation[i]);
+    const proposedFunds = bigNumberify(proposedAllocation[i]);
+    const topUp = currentFunds.gte(proposedFunds)
+      ? '0x0'
+      : proposedFunds.sub(currentFunds).toHexString();
+    newAllocation[i] = topUp;
+  }
+  return newAllocation;
+}
```
andrewgordstewart: I feel like the accounting logic is complicated enough to merit unit testing.

packages/wallet/src/redux/protocols/concluding/responder/reducer.ts
```diff
@@ -181,14 +229,52 @@ function concludeApproved(protocolState: NonTerminalCState, sharedData: Storage)
       protocolState.channelId,
     );
     return {
-      protocolState: decideDefund({ ...protocolState }),
+      protocolState: decideDefund({ ...protocolState, opponentSelectedKeepLedgerChannel: false }),
```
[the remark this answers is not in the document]
lalexgap: This means that we have not heard from your opponent's wallet.
The two options are 
- Opponent decides to defund, which automatically transitions you into `defunding`
- Opponent decides to keep ledger channel open. This sets `opponentSelectedKeepLedgerChannel` to `true`.

We use `opponentSelectedKeepLedgerChannel` to know if we can transition directly to `consensus-update` or if we're waiting on our opponent still.

Maybe a better name would be `opponentHasSelected` ?

packages/rps/src/redux/message-service/saga.ts
```diff
@@ -390,3 +394,7 @@ function* signMessage(commitment: Commitment) {
     }
   }
 }
+
+function sanitizeMessageForFirebase(message) {
+  return JSON.parse(JSON.stringify(message));
```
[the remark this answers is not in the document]
lalexgap: Firebase cannot handle `undefined` values so this sanitizes them out.

packages/wallet/src/redux/protocols/concluding/responder/reducer.ts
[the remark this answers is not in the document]
andrewgordstewart: That might be a bit more clear, yes!

packages/wallet/src/redux/protocols/concluding/responder/readme.md
```diff
@@ -29,18 +29,23 @@ linkStyle default interpolate basis
   MT  --> |No| AF(ResponderAcknowledgeFailure)
   CC  --> |WALLET.CONCLUDING.RESPONDER.CONCLUDE.APPROVED| DD(ResponderDecideDefund)
   DD --> |WALLET.CONCLUDING.RESPONDER.DEFUND.CHOSEN| D(ResponderWaitForDefund)
+  DD --> |WALLET.CONCLUDING.RESPONDER.KEEP_OPEN.CHOSEN| WO(WaitForOpponentDecision)
   DD --> |COMMITMENT_RECEIVED| D(ResponderWaitForDefund)
-  D   --> |defunding protocol succeeded| AS(ResponderAcknowledgeSuccess)
+  WO --> |COMMITMENT_RECEIVED| D(ResponderWaitForDefund)
+  WO -->|WALLET.CONCLUDING.KEEP_OPEN_SELECTED|CU(WaitForLedgerUpdate)
+  CU -->|COMMITMENT_RECEIVED|CU
```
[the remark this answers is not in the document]
lalexgap: If the instigator chooses to defund they just send their defund commitment which transitions us to `ResponderWaitForDefund`.

These are the two transitions that handle the instigator choosing to defund:
 ```DD --> |COMMITMENT_RECEIVED| D(ResponderWaitForDefund)
  D   --> |defunding protocol succeeded| AS(ResponderAcknowledgeSuccess)	  WO --> |COMMITMENT_RECEIVED| D(ResponderWaitForDefund)```

packages/wallet/src/redux/protocols/consensus-update/reducer.ts
```diff
@@ -0,0 +1,114 @@
...
+      `Ledger Top Up Protocol expected COMMITMENT_RECEIVED received ${action.type} instead.`,
+    );
+    return { protocolState, sharedData };
+  }
+  if (protocolState.type !== 'ConsensusUpdate.WaitForUpdate') {
+    console.warn(`Consensus update reducer was called with terminal state ${protocolState.type}`);
+    return { protocolState, sharedData };
+  }
+
+  const checkResult = checkAndStore(sharedData, action.signedCommitment);
```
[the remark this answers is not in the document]
lalexgap: Yeah. I just got the essential bits working for my purposes but that check is easy to add so I'll throw it in.

packages/wallet/src/redux/protocols/direct-funding/__tests__/reducer.test.ts
```diff
@@ -32,6 +32,27 @@ describe('Player A Happy path', () => {
     itTransitionsTo(updatedState, 'DirectFunding.FundingSuccess');
   });
 });
+describe('Player A No Exchanging Post Fund Setups Happy path', () => {
```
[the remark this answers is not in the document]
lalexgap: Probably would be better titled `Player A Happy Path With No Post Fund Setup Exchange`

packages/wallet/src/redux/protocols/direct-funding/actions.ts
lalexgap: The main change to the DirectFunding reducer was adding a exchangePostFundSetups` flag that can be set to determine whether the DirectFunding reducer sends Post Fund Setups after funding.

When topping up an existing ledger channel we don't want to send post fund setups for the ledger channel, we just want to deposit into the channel and then exchange updates after. However before these changes the direct funding reducer would automatically send postFundSetups for the ledger channel after deposit.

I added the `exchangePostFundSetup` flag so we can use the `DirectFunding` reducer from the `ledger-topup` protocol without breaking existing protocols.

packages/wallet/src/redux/protocols/existing-channel-funding/reducer.ts
lalexgap: This check is determine if either player needs to top up their ledger channel. If either player needs a top-up both players start the `ledger-topup` protocol which figures out who needs to add more funds and exchanges `commitment`s as necessary.

So in the case of `[A:4,B:6]` and proposedAmount of `10`.  This would return true for both players since `A <10/2`. However in the `ledger-topup` protocol only A would be required to add more funds.

packages/wallet/src/redux/protocols/funding/player-b/reducer.ts
lalexgap: Yeah, I think there's more work to be done around this. There should be a negotiation between two players on whether they have the existing ledger channel and if they want to use it.

packages/wallet/src/redux/protocols/ledger-top-up/reducer.ts
lalexgap: I've created https://github.com/magmo/apps/issues/543 to address this in a separate PR. I thought I was being clever :(

packages/wallet/src/redux/protocols/ledger-top-up/reducer.ts
lalexgap: That's a good point. I should really be going off the `destination`.

packages/wallet/src/redux/protocols/ledger-top-up/reducer.ts
lalexgap: It's just a local helper function but it probably does make sense to share it in `reducer-helpers`

packages/wallet/src/redux/protocols/existing-channel-funding/reducer.ts
lalexgap: I was debating that as well. That probably makes more sense in hindsight.

packages/wallet/src/redux/protocols/ledger-top-up/reducer.ts
lalexgap: Deferring until #543

packages/wallet/src/redux/protocols/consensus-update/__tests__/scenarios.ts
```diff
@@ -0,0 +1,127 @@
...
+import { CONSENSUS_UPDATE_PROTOCOL_LOCATOR } from '../reducer';
+
+const twoThree = [
+  { address: asAddress, wei: bigNumberify(2).toHexString() },
+  { address: bsAddress, wei: bigNumberify(3).toHexString() },
+];
+const twoThreeOneTwo = [
+  { address: asAddress, wei: bigNumberify(2).toHexString() },
+  { address: bsAddress, wei: bigNumberify(3).toHexString() },
+  { address: asAddress, wei: bigNumberify(1).toHexString() },
```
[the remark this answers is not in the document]
lalexgap: See #543

packages/wallet/src/redux/protocols/ledger-top-up/reducer.ts
lalexgap: This can be addressed in #543

packages/wallet/src/redux/protocols/existing-channel-funding/reducer.ts
andrewgordstewart: But it assumes that each player wants to have the same amount of funds staked in the channel. What if we have a one-way relationship, where I pay you, and I want to top up my end of the channel?

lalexgap: At that's right. I've created #545 to address this.

PR #512, opened 2019-05-26
Ledger challenging

### LINKED ISSUE #503: Enable challenging in the indirect-defunding protocol
In case your opponent has rage-quit and you need to get your money back.

geoknee:
Updating the `indirect-defunding` protocol to allow for challenges. Closes #503.

Screenshot 2019-05-26 at 14 40 10

I am taking an approach that attempts to reuse the (top level) dispute protocol. On the one hand, this may pose some problems with action routing / container routing. On the other hand, we probably need to make this work to maintain our modular approach.

Updates to 
- [x] Readme
- [x] States
- [x] Actions
- [x] Scenarios (happy path)
- [x] Tests (happy path)
- [x] Reducer (happy path)
- [x] Container / Components (happy path)

- [ ] Scenarios (unhappy paths)
- [ ] Tests (unhappy paths)
- [ ] Reducer (unhappy paths)
- [ ] Container / Components (unhappy paths)


geoknee:
Closing this since I was getting TS errors when trying to embed `dispute` state on `indirect-defunding` state (e.g. circularly defined unions, because `indirect-defunding` itself can be embedded in `dispute` state). Decided to pursue an alternative approach where a new dispute process spins up (instead of being embedded).

PR #515, opened 2019-05-28
Used one address per a wallet

lalexgap:
To allow the reuse of ledger channels we want to keep the address consistent between channels.

We now generate an address when the wallet is initialized and just increment the ledger nonce(instead of using a nonce of 0) instead of generating an address per channel.


packages/wallet/src/__stories__/dummy-wallet-states.ts
```diff
@@ -47,6 +47,8 @@ const defaults = {
   uid: 'uid',
   processStore: {},
```
andrewgordstewart: The surrounding object strikes me as something that could be exported from some test file, and referenced elsewhere.

packages/wallet/src/redux/selectors.ts
```diff
@@ -67,3 +67,24 @@ export const getProtocolForProcessId = (
 export const getProtocolState = (state: walletStates.Initialized, processId: string) => {
   return state.processStore[processId].protocolState;
 };
+
+export const getLatestNonce = (
+  state: SharedData,
+  participantA: string,
+  participantB: string,
+  libraryAddress: string,
```
andrewgordstewart: What about just using the latest nonce for the given `libraryAddress`?

packages/wallet/src/redux/protocols/indirect-funding/player-a/reducer.ts
```diff
@@ -275,15 +276,24 @@ function handleWaitForDirectFunding(
...
   const appAttributes = {
     proposedAllocation: [],
     proposedDestination: [],
     furtherVotesRequired: 0,
     updateType: UpdateType.Consensus,
   };
-  // TODO: We'll run into collisions if we reuse the same nonce
-  const nonce = 0;
+
+  const nonce = selectors.getLatestNonce(
```
andrewgordstewart: This has the effect of re-using the latest nonce, a nonce that has already been used.

Why not have a selector called `getNextNonce` to make this mistake more obvious?

packages/wallet/src/redux/selectors.ts
lalexgap: We can do that. It just means we burn through nonces faster.

packages/wallet/src/__stories__/dummy-wallet-states.ts
lalexgap: I'm not sure we'd want to export an empty wallet state that has faked values in it. We could probably switch stories to use some our tests states but I think that's beyond the scope of this PR.

PR #519, opened 2019-05-29
Remove private field from wallet package.json

lalexgap:
Remove private field from wallet package


lalexgap:
> This looks good. But why is `magmo-wallet` a private package?

I think the idea is that we can publish `magmo-wallet` as a private package since it's not really in a ready state for others to consume. Although I'm not sure if its worth the extra complication versus publishing it publicly?

kerzhner:
I would argue that all of our software is alpha at this point. It would be nice to be consistent and either create just private packages or just public packages.

lalexgap:
> I would argue that all of our software is alpha at this point. It would be nice to be consistent and either create just private packages or just public packages. That being said, I do not feel strongly on this matter.

I think I'd tend to agree with you. @andrewgordstewart any strong opinions?

andrewgordstewart:
@lalexgap No strong opinions from me, as long as we avoid `v1.0`

PR #520, opened 2019-06-03
Ledger disputes during indirect-defunding

### LINKED ISSUE #503: Enable challenging in the indirect-defunding protocol

geoknee:
Updating the indirect-defunding protocol to allow for challenges. Closes #503. In contrast to #512, here a new `Dispute-` process is started when the challenge is detected. The dispute process then needs to 'hand back' to the `indirect-defunding` protocol in a similar way that it currently 'hands back' to the RPS game. 

TODO:
- [x] tidy up the changes (type safety, naming, removing unecessary display messages, etc)
- [x] handle happy path when `indirect-defunding` protocol initialized
- [x] handle happy path when `indirect-defunding` _not_ initialized.
- [x] write tests for new behaviour (challenger)
- [x] write tests for new behaviour (responder)
- [x] write tests for new behaviour (indirect-defunding)
- [x] write tests for new behaviour (concluding)
- [ ] handle ledger challenge timeouts


geoknee:
Expiring ledger challenges pose some challenges for withdrawing, that we should discuss possible solutions for (ass well as deciding on whether those solutions could be implemented in a new PR.)

There are the following four scenarios (A and B are participants, X is an app channel, L is a ledger channel):

1. X concluded OFF chain, L finalized as funding X
2. X concluded OFF chain, L finalized as funding A,B
3. X finalized ON chain, L finalized as funding X
4. X finalized ON chain, L finalized as funding A,B

with corresponding withdrawal strategies: 

- [ ] 1. register conclusion proof for X on chain, transfer from L to X, transfer from X to A/B, withdraw
- [x] 2. transfer from L to A/B, withdraw 
- [ ] 3. transfer from L to X, transfer from X to A/B, withdraw
- [x] 4. transfer from L to A/B, withdraw

Strategies 1 and 3 require either i) multiple transactions or ii) new functions in the adjudicator contract that combine the on chain operations accordingly. The wallet currently applies strategy 2/4 in all the above scenarios, and hence we get a tx error in scenario 1, and some unexpected behaviour in 3 (where A can withdraw all off the funds and B cannot withdraw any funds from L, despite it funding X only). Scenarios 2 and 4 play out correctly.

PR #522, opened 2019-06-06
Making the adjudicator watcher test synchronous

kerzhner:
Hoping to address this Circle failure: https://circleci.com/gh/magmo/apps/1965


kerzhner:
You bring up good points. We can still merge in this PR as the modified test case does not need to be asynchronous. So the PR makes it easier to think about how the test case is executed.

andrewgordstewart:
Is it true that the test actually runs the expectations before `it` returns? That's the biggest part that I'm worried about

kerzhner:
In this case, the test runner waits until the `done` function is executed.

PR #523, opened 2019-06-12
Ensure destination field does not change in application channels

geoknee:
The `destination` field was introduced with Nitro (to allow the a channel to allocate to non-participant addresses) -- but our apps are only safe under the old ForceMove protocol since no checks on `destination` are performed inside `validTransition`. There is an attack where I overwrite `destination` with my own address as the single beneficiary, and I can then transfer more funds than I am entitled to into my own account.


geoknee:
TicTacToe still uses the simple adjudicator, so no changes required just now.

PR #525, opened 2019-06-13
Allow `Application-` to continue when instigator cancels conclude

geoknee:
Previously, the mere request for a conclude was wrapping up the application process. Since the instigator can back out, this is undesired behaviour because the application process needs to continue.

Instead, it should 
- [x] not be wrapped up on a conclude request

and only be wrapped up when 

- [x] the app channel is registered finalized on chain
- [x] a conclude commitment is sent or received (i.e. channel finalized off chain)

~~As far as I can tell, conclude commitments are not verified (at least not in the same way as regular commitments). There is an open question about exactly where and how to dispatch the action that will wrap up the application process, when the channel is finalized off chain.~~

Message sender and listener sagas will inspect the messages, and conditionally put a new `concluded` action, to be routed to the `Application` process, as appropriate.

### DIFF
--- a/packages/wallet/src/redux/protocols/application/actions.ts
+++ b/packages/wallet/src/redux/protocols/application/actions.ts
@@ -17,8 +17,8 @@ export interface OpponentCommitmentReceived {
   commitment: Commitment;
   signature: string;
 }
-export interface ConcludeRequested {
-  type: 'WALLET.APPLICATION.CONCLUDE_REQUESTED';
+export interface Concluded {
+  type: 'WALLET.APPLICATION.CONCLUDED';
   processId: string;
 }
 
@@ -45,10 +45,10 @@ export const opponentCommitmentReceived: ActionConstructor<OpponentCommitmentRec
   };
 };
 
-export const concludeRequested: ActionConstructor<ConcludeRequested> = p => {
+export const concluded: ActionConstructor<Concluded> = p => {
   const { processId } = p;
   return {
-    type: 'WALLET.APPLICATION.CONCLUDE_REQUESTED',
+    type: 'WALLET.APPLICATION.CONCLUDED',
     processId,
   };
 };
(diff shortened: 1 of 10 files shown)


packages/wallet/src/redux/protocols/application/actions.ts
```diff
@@ -17,8 +17,8 @@ export interface OpponentCommitmentReceived {
   commitment: Commitment;
   signature: string;
 }
-export interface ConcludeRequested {
-  type: 'WALLET.APPLICATION.CONCLUDE_REQUESTED';
+export interface Concluded {
```
tomclose: I'm not sure I understand the name change here. `ConcludeRequested` seems more descriptive of what's actually happening, unless I've misunderstood?

packages/wallet/src/redux/sagas/message-sender.ts
```diff
@@ -1,9 +1,18 @@
 import { put } from 'redux-saga/effects';
 import { messageSent } from '../actions';
+import { concluded } from '../protocols/application/actions';
+import { APPLICATION_PROCESS_ID } from '../protocols/application/reducer';
 
 export function* messageSender(message) {
   window.parent.postMessage(message, '*');
-
+  console.log(message);
```
tomclose: ✂️

packages/wallet/src/redux/protocols/application/actions.ts
geoknee: 'WALLET.NEW_PROCESS.CONCLUDE_REQUESTED' still exists, but the point is that it spins up a protocol which can *fail* (by the user backing out / cancelling). The `requested` part seems apt for this action.

From the application reducer's point of view, we don't want to do anything when the user requests a conclude; we only want to terminate and stop signing commitments once we know for sure that the channel is `Concluded`.

PR #528, opened 2019-06-18
AdvanceChannel protocol

andrewgordstewart:


packages/wallet/src/redux/protocols/advance-channel/readme.md
```diff
@@ -0,0 +1,65 @@
+### Advancing the stage a channel
+
+The `AdvanceStage` protocol accepts two mandatory inputs and one optional input:
```
lalexgap: Do we expect to use this to advance from app-> conclude? If we're planning on using this just for `setup` commitments it might make sense to include `setup` in the name and only accept `CommitmentType.PreFundSetup` and `CommitmentType.PostFundSetup`

andrewgordstewart: I couldn't see a reason why it couldn't be used when concluding the game, so I would like it to be used there if possible as well.

PR #530, opened 2019-06-19
Un nest defunding from dispute

### LINKED ISSUE #529: De-nest defunding from under dispute

geoknee:
This PR is part of a larger project which will make it possible to embed the `dispute` protocol underneath other protocols, without running into circular definitions in `DisputeAction` (which otherwise contains all of the `DefundingAction`s, and therefore all the `IndirectDefundingAction`s, and therefore all the `DisputeAction`s if we want to embed it there).

Closes #529 .

- [x] readmes
- [x] states
- [x] actions
- [x] container
- [x] stories
- [x] scenarios
- [x] reducers
- [x] spin up defunding as a new process on challenge timeout
- [x] tests
- [x] test new behaviour
- [x] handle commitments arriving before process exists

My solution to a commitment arriving before a process exists: the player who sends the commitment *first* relays a whitelisted `NEW_PROCESS.-` action, so that the recipient has that process in place and ready for the commitment (which follows immediately) to be routed correctly. 

This involves queuing up 2 messages from the wallet (which works great), but required a small tweak to the rps saga to actually post these to firebase properly.

Screenshot 2019-06-21 at 14 45 03

has become

Screenshot 2019-06-21 at 15 44 48

Further work for separate PRs:

- unnest defunding from concluding (seems ugly to have a top level protocol parenting another top level protocol)
- nest all top level protocols under `application`, which can manage spawning new processes when others terminate


packages/wallet/src/redux/protocols/indirect-defunding/reducer.ts
```diff
@@ -44,14 +46,22 @@ export const initialize = (
...
     }
     newSharedData = signResult.store;
+    // send a request for opponent to start new defunding process first, because they may not yet have done so
+
+    const actionToRelay = defundRequested({
+      processId: '0XTODO', // wish to terminate existing process in opponent's wallet (could use currentProcessId but this is not on SharedData)
+      channelId,
+    });
+
+    const messageRelay0 = messageRelayRequested(theirAddress(ledgerChannel), actionToRelay);
```
lalexgap: could we rename `messageRelay0` to something more meaningful like `defundMessageRelay`?

packages/wallet/src/redux/reducer.ts
```diff
@@ -51,6 +52,11 @@ export function initializedReducer(
     newState = updateSharedData(newState, action);
   }
 
+  // if the action is both a protocol action and a new process action (e.g. action terminates current process and starts new one)
+  if (isNewProcessAction(action) && actions.isProtocolAction(action)) {
```
lalexgap: Instead of having an `action` that creates a new process and terminates an old one would it make sense to just have two actions? IE (`ExitChallenge` and `DefundRequested`).  That way we can keep the distinction between new process actions and protocol actions.

packages/wallet/src/redux/protocols/indirect-defunding/reducer.ts
```diff
@@ -44,14 +46,22 @@ export const initialize = (
       };
     }
     newSharedData = signResult.store;
+    // send a request for opponent to start new defunding process first, because they may not yet have done so
+
+    const actionToRelay = defundRequested({
+      processId: '0XTODO', // wish to terminate existing process in opponent's wallet (could use currentProcessId but this is not on SharedData)
```
lalexgap: Do we expect there to be an opponent's process that would be terminated here or would that already be handled by the challenging reducers?

packages/rps/src/redux/message-service/saga.ts
```diff
@@ -66,7 +66,7 @@ export function* sendWalletMessageSaga() {
         console.error(err);
       }
     } else {
-      yield call(reduxSagaFirebase.database.create, `/messages/${to.toLowerCase()}`, messageToSend);
+      yield fork(reduxSagaFirebase.database.create, `/messages/${to.toLowerCase()}`, messageToSend);
```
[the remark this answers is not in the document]
geoknee: With `call`, the saga is blocked on the firebase API call completing, and so the next `MESSAGE_RELAY_REQUESTED` can be missed. See  the section entitled 'My saga is missing dispatched actions` at https://redux-saga.js.org/docs/Troubleshooting.html . Using `fork` makes it a non-blocking call.

packages/wallet/src/redux/reducer.ts
geoknee: In principle yes, but that would seem to steer us toward an internal message queue (something that we may not wish to have).

packages/wallet/src/redux/protocols/indirect-defunding/reducer.ts
geoknee: The opponent might still be running that process: for example, they are at the final stage of `Dispute`, and haven't yet decided whether to defund or not.

packages/wallet/src/redux/reducer.ts
geoknee: If (in future work) these protocols are nested under `Application-`, they won't need to be spawned as new processes, and this action can just be a `ProtocolAction`.

packages/wallet/src/redux/protocols/indirect-defunding/reducer.ts
geoknee: :+1: addressed in https://github.com/magmo/apps/pull/530/commits/4e1fc8a2317d5098fed7a54cc0efc5b71455fb6a

packages/wallet/src/redux/protocols/indirect-defunding/reducer.ts
geoknee: Have moved `currentProcessId?` to sharedData and substituted it for '0xTODO` here (https://github.com/magmo/apps/pull/530/commits/4e1fc8a2317d5098fed7a54cc0efc5b71455fb6a)

PR #531, opened 2019-06-19
Store arbitrary number of commitments on channelState

andrewgordstewart:
This feature is required to implement virtual channels.

Some commits unrelated to this feature are included for my convenience, which make a little progress towards the AdvanceChannel protocol.

### DIFF
--- a/packages/wallet/src/redux/protocols/dispute/challenger/__tests__/scenarios.ts
+++ b/packages/wallet/src/redux/protocols/dispute/challenger/__tests__/scenarios.ts
@@ -4,10 +4,7 @@ import * as tsScenarios from '../../../transaction-submission/__tests__';
 import { setChannel, EMPTY_SHARED_DATA } from '../../../../state';
 import { ChannelState } from '../../../../channel-store';
 import * as channelScenarios from '../../../../__tests__/test-scenarios';
-import {
-  channelFromCommitments,
-  partiallyOpenChannelFromCommitment,
-} from '../../../../channel-store/channel-state/__tests__';
+import { channelFromCommitments } from '../../../../channel-store/channel-state/__tests__';
 import {
   challengeExpiredEvent,
   respondWithMoveEvent,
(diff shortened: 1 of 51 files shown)


packages/wallet/src/redux/types.ts
```diff
@@ -1,4 +1,10 @@
-export const enum PlayerIndex {
+export const enum TwoPartyPlayerIndex {
   A = 0,
   B = 1,
 }
+
+export const enum ThreePartyPlayerIndex {
```
kerzhner: Do we foresee any cases with 4 parties coming up? Should we start moving away from two and three party cases toward n-party structures? Or is the added complexity not worth it?

packages/wallet/src/redux/protocols/reducer-helpers.ts
```diff
@@ -45,18 +45,13 @@ export const filterOutSignatureMessages = (sideEffects?: SideEffects): SideEffec
 
 export function sendFundingComplete(sharedData: SharedData, appChannelId: string) {
   const channelState = selectors.getOpenedChannelState(sharedData, appChannelId);
-  const lastCommitment = channelState.lastCommitment.commitment;
-  if (
-    lastCommitment.commitmentType !== CommitmentType.PostFundSetup ||
-    lastCommitment.turnNum !== 3
-  ) {
+  const c = getLastCommitment(channelState);
```
kerzhner: Not a fan of one letter variable names.

packages/wallet/src/redux/protocols/dispute/challenger/__tests__/scenarios.ts
```diff
@@ -32,14 +29,18 @@ const {
   signedCommitment21,
 } = channelScenarios;
 
-const partiallyOpen = partiallyOpenChannelFromCommitment(signedCommitment0, address, privateKey);
+const partiallyOpen = channelFromCommitments([signedCommitment0], address, privateKey);
 const theirTurn = channelFromCommitments(
-  signedCommitment19,
-  signedCommitment20,
+  [signedCommitment19, signedCommitment20],
+
```
kerzhner:
```suggestion
```

packages/wallet/src/redux/types.ts
andrewgordstewart: I don't know if we'll ever get rid of all of the reliance on two parties. If we do, then we should get rid of the `*PlayerIndex` types, for sure.

PR #532, opened 2019-06-19
WIP: Initializing the advance-channel protocol

andrewgordstewart:
I think that the initialization logic is complete, but it is not yet tested for the hub. I am adding tests for the scenarios in order.

### DIFF
--- a/packages/wallet/src/redux/__tests__/helpers.ts
+++ b/packages/wallet/src/redux/__tests__/helpers.ts
@@ -167,3 +167,10 @@ export const itStoresThisCommitment = (
     expect(lastSignedCommitment).toMatchObject(signedCommitment);
   });
 };
+
+export const itRegistersThisChannel = (state: SharedData, channelId: string, processId: string) => {
+  it('subscribes to channel events in the channel subscriptions', () => {
+    const subscriptionState = state.channelSubscriptions[processId];
+    expect(subscriptionState).toContain(channelId);
+  });
+};
--- a/packages/wallet/src/redux/__tests__/helpers.ts
+++ b/packages/wallet/src/redux/__tests__/helpers.ts
@@ -79,8 +79,34 @@ export const expectThisMessageAndCommitmentSent = (
     expect(item.messagePayload.signedCommitment.commitment).toMatchObject(c);
   });
 };
+
+export const expectThisMessageAndTheseCommitmentsSent = (
+  state: SideEffectState,
+  commitments: PartialCommitments,
+  messageType: string,
+) => {
+  expectSideEffect('messageOutbox', state, item => {
+    expect(item.messagePayload.type).toEqual(messageType);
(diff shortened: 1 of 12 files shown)


packages/wallet/src/redux/protocols/advance-channel/__tests__/reducer.test.ts
```diff
@@ -0,0 +1,39 @@
...
+) => {
+  it(`transitions to ${type}`, () => {
+    expect(result.type).toEqual(type);
+  });
+};
+
+describe('sending preFundSetup as A', () => {
+  const scenario = scenarios.newChannelAsA;
+
+  describe.only('when initializing', () => {
```
andrewgordstewart: The tests are not fully written, but I think the scenarios are mostly complete.

Only this test is enabled.

packages/wallet/src/redux/protocols/advance-channel/__tests__/scenarios.ts
```diff
@@ -53,6 +47,18 @@ const propsB = {
   ourIndex: ThreePartyPlayerIndex.B,
 };
 
+const propsHub = {
+  ...props,
+  ourIndex: ThreePartyPlayerIndex.Hub,
+};
+
+const commitments0 = [signedCommitment0];
```
kerzhner: Do you think there is a more descriptive yet succinct variable name that describes the channel state for the sequence of commitments?

packages/wallet/src/redux/__tests__/helpers.ts
```diff
@@ -79,9 +79,35 @@ export const expectThisMessageAndCommitmentSent = (
...
+
+export const expectThisMessageAndTheseCommitmentsSent = (
+  state: SideEffectState,
+  commitments: PartialCommitments,
+  messageType: string,
+) => {
+  expectSideEffect('messageOutbox', state, item => {
+    expect(item.messagePayload.type).toEqual(messageType);
+    expect(item.messagePayload.signedCommitments).toMatchObject(
+      commitments.map(({ commitment }) => ({ commitment })), // This has the effect of ignoring the signature
```
kerzhner: Can't wrap my head around this...

andrewgordstewart: `commitments` are signed (partial) commitments, so they have the shape `{ commitment: Commitment, signature: string }`

This helper is only concerned with the commitment part of it, and doesn't assert anything about the signature (in keeping with the existing helper for a single commitment)

packages/wallet/src/redux/protocols/advance-channel/__tests__/scenarios.ts
andrewgordstewart: Perhaps, but I'm far ahead of origin's HEAD locally, so I would prefer not to rename it in this PR.

PR #533, opened 2019-06-20
WIP: Advance channel reducer

andrewgordstewart:
All PreFundSetup tests pass for all three players.

I believe the entire logic is now almost correct, aside from the missing `initializeWithExistingChannel`.

### DIFF
--- a/packages/wallet/src/redux/protocols/advance-channel/__tests__/reducer.test.ts
+++ b/packages/wallet/src/redux/protocols/advance-channel/__tests__/reducer.test.ts
@@ -55,3 +55,32 @@ describe('sending preFundSetup as A', () => {
     itStoresThisCommitment(result, commitments[2]);
   });
 });
+
+describe('sending preFundSetup as B', () => {
+  const scenario = scenarios.newChannelAsB;
+  const { processId } = scenario;
+
+  describe('when initializing', () => {
+    const { sharedData, args } = scenario.initialize;
+    const { protocolState, sharedData: result } = initialize(
+      processId,
+      sharedData,
+      CommitmentType.PreFundSetup,
+      args,
+    );
+
+    itTransitionsTo(protocolState, 'AdvanceChannel.ChannelUnknown');
+    itSendsNoMessage(result);
+  });
+
+  describe.only('when receiving prefund commitments from A', () => {
+    const { state, sharedData, action, commitments } = scenario.receiveFromA;
+
+    const { protocolState, sharedData: result } = reducer(state, sharedData, action);
+
+    itTransitionsTo(protocolState, 'AdvanceChannel.CommitmentSent');
+    itStoresThisCommitment(result, commitments[1]);
+    // expectTheseCommitmentsSent(result, commitments);
+    // itRegistersThisChannel(result, channelId, processId);
+  });
+});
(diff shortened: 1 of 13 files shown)


packages/wallet/src/redux/protocols/advance-channel/readme.md
```diff
@@ -2,11 +2,12 @@
```
kerzhner: Do you mean "Advancing the stage of a channel" for the title?

packages/wallet/src/redux/protocols/advance-channel/__tests__/reducer.test.ts
```diff
@@ -43,15 +43,81 @@ describe('sending preFundSetup as A', () => {
...
     const { state, sharedData, action, commitments } = scenario.receiveFromHub;
+    const { protocolState, sharedData: result } = reducer(state, sharedData, action);
+
+    itTransitionsTo(protocolState, 'AdvanceChannel.Success');
+    itStoresThisCommitment(result, commitments[2]);
+    itSendsNoMessage(result);
+  });
+});
+
+describe('sending preFundSetup as Hub', () => {
```
kerzhner: Seems like we are on the hook for at least _trying_ to use this wallet for the hub now 😉

packages/wallet/src/redux/protocols/advance-channel/reducer.ts
```diff
@@ -147,57 +145,175 @@ function initializeWithNewChannel(
...
+const channelUnknownReducer: ProtocolReducer<states.AdvanceChannelState> = (
+  protocolState: states.ChannelUnknown,
+  sharedData,
+  action: CommitmentsReceived,
+) => {
+  const { ourIndex, privateKey, commitmentType } = protocolState;
+  const channelId = getChannelId(action.signedCommitments[0].commitment);
+
+  const checkResult = checkAndInitialize(sharedData, action.signedCommitments[0], privateKey);
+  if (!checkResult.isSuccess) {
```
kerzhner: Will the stack trace be easier to interpret if the error is thrown next to the logic where the failure occurs?

packages/wallet/src/redux/protocols/advance-channel/reducer.ts
```diff
@@ -147,57 +145,175 @@ function initializeWithNewChannel(
...
+    throw new Error('Could not initialize channel');
+  }
+  sharedData = checkResult.store;
+  sharedData = checkCommitments(sharedData, 0, action.signedCommitments);
+  let channel = getChannel(sharedData.channelStore, channelId);
+  if (!channel) {
```
kerzhner: Can we remove this check if we have sufficient unit testing?

andrewgordstewart: Unfortunately not, without refactoring `getChannel` -- it can return `undefined`.

@lalexgap do you think we could throw in `getChannel` if it doesn't exist?

lalexgap: I think that would make sense.

packages/wallet/src/redux/protocols/advance-channel/readme.md
andrewgordstewart: Fixing upstream.

PR #535, opened 2019-06-20
Clear the Library folder on build

### LINKED ISSUE #534: Wallet Build Fails if `lib` is not empty
If the `lib` folder is not empty than the wallet build command fails with these errors: ```error TS5055: Cannot write file '/Users/alexgap/code/apps/packages/wallet/lib/src/redux/protocols/withdrawing/states.d.ts' because it would overwrite input file.

error TS5055: Cannot write file '/Users/alexgap/code/apps/packages/wallet/lib/src/redux/state.d.ts' because it would overwrite input file.

error TS5055: Cannot write file '/Users/alexgap/code/apps/packages/wallet/lib/src/redux/types.d.ts' because it would overwrite input file.

error TS5055: Cannot write file '/Users/alexgap/code/apps/packages/wallet/lib/src/redux/utils.d.ts' because it would overwrite input file.

error TS5055: Cannot write file '/Users/alexgap/code/apps/packages/wallet/lib/src/utils/hex-utils.d.ts' because it would overwrite input file.

error TS5055: Cannot write file '/Users/alexgap/code/apps/packages/wallet/lib/src/utils/reducer-utils.d.ts' because it would overwrite input file.

Found 76 errors.```

lalexgap:
This fixes #534.

 I spent some time trying to figure out why the `.d.ts` files were being included from the `lib` folder but didn't manage to figure it out.

This avoids the problem by clearing the `lib` folder before building.

### DIFF
--- a/packages/wallet/package.json
+++ b/packages/wallet/package.json
@@ -9,7 +9,7 @@
   "scripts": {
     "start": "npm run clearContracts && node scripts/start.js",
     "build": "run-s clearContracts build:typescript build:webpack",
-    "build:typescript": "npx tsc",
+    "build:typescript": "yarn run clearLib && npx tsc",
     "build:webpack": "node scripts/build.js",
     "prepare": "npx tsc",
     "test": "run-s prettier:check 'test:app --all'",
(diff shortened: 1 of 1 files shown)


lalexgap:
This seems to only occur on #509 so I'm closing this.

PR #538, opened 2019-06-21
Advance channel reducer when channel exists

andrewgordstewart:


packages/wallet/src/redux/protocols/advance-channel/reducer.ts
```diff
@@ -201,6 +234,33 @@ const channelUnknownReducer: ProtocolReducer<states.AdvanceChannelState> = (
...
+  const checkResult = checkAndInitialize(sharedData, action.signedCommitments[0], privateKey);
+  if (!checkResult.isSuccess) {
+    throw new Error('Could not initialize channel');
+  }
+  sharedData = checkResult.store;
+  sharedData = checkCommitments(sharedData, 0, action.signedCommitments);
+
+  const result = attemptToAdvanceChannel(sharedData, protocolState, channelId);
+  sharedData = result.sharedData;
+  const nextProtocolState = result.protocolState; // The type might have changed, so we can't overwrite protocolState
```
kerzhner: It doesn't seem like `protolState` assignment is a good idea anyways...

packages/wallet/src/redux/protocols/advance-channel/reducer.ts
```diff
@@ -149,22 +147,60 @@ function initializeWithNewChannel(
...
-  const channelId = getChannelId(action.signedCommitments[0].commitment);
+function initializeWithExistingChannel(
+  processId,
+  sharedData: Storage,
+  initializeChannelArgs: OngoingChannelArgs,
+) {
+  const { channelId, ourIndex } = initializeChannelArgs;
+  const channel = getChannel(sharedData.channelStore, channelId);
+  if (!channel) {
+    throw new Error('Channel not stored');
```
kerzhner: Should we standardize this error for the scenario of channel expected but missing?

packages/wallet/src/redux/protocols/advance-channel/reducer.ts
```diff
@@ -149,22 +147,60 @@ function initializeWithNewChannel(
...
+  }
+  if (isSafeToSend({ sharedData, ourIndex })) {
+    const lastCommitment = getLastCommitment(channel);
+    const ourCommitment = nextSetupCommitment(lastCommitment);
+    if (ourCommitment === 'NotASetupCommitment') {
```
kerzhner: Is a commitment object being compared to a string?

packages/wallet/src/redux/protocols/advance-channel/reducer.ts
```diff
@@ -201,6 +234,33 @@ const channelUnknownReducer: ProtocolReducer<states.AdvanceChannelState> = (
...
+  const { privateKey } = protocolState;
+  const channelId = getChannelId(action.signedCommitments[0].commitment);
+  const checkResult = checkAndInitialize(sharedData, action.signedCommitments[0], privateKey);
+  if (!checkResult.isSuccess) {
+    throw new Error('Could not initialize channel');
+  }
+  sharedData = checkResult.store;
+  sharedData = checkCommitments(sharedData, 0, action.signedCommitments);
+
+  const result = attemptToAdvanceChannel(sharedData, protocolState, channelId);
```
kerzhner: Should we use destructuring assignment here?

packages/wallet/src/redux/protocols/advance-channel/reducer.ts
andrewgordstewart: `nextSetupCommitment` currently returns `Commitment | 'NotASetupCommitment'`, so this check narrows the type to `Commitment`

packages/wallet/src/redux/protocols/advance-channel/reducer.ts
andrewgordstewart: Destructuring is pretty awkward here:
- You can't overwrite `protocolState` because it might change the type
- You can't re-declare `sharedData` as a const
- If you declare a new `nextSharedData` variable, you have to then assign `sharedData = nextSharedData`, or do something else, to take into account that you might then register the channel to monitor on L260

packages/wallet/src/redux/protocols/advance-channel/reducer.ts
andrewgordstewart: This check isn't necessary, now that `getChannel` always returns a `ChannelState`.

PR #540, opened 2019-06-21
Refactor direct funding to use AdvanceChannel

andrewgordstewart:


packages/wallet/src/redux/protocols/direct-funding/__tests__/scenarios.ts
```diff
@@ -73,78 +51,68 @@ const bFundingReceivedEvent = globalActions.fundingReceivedEvent({
...
-const postFundSetup0 = globalActions.commitmentReceived({
-  processId: channelId,
-  signedCommitment: scenarios.signedCommitment2,
-});
-const postFundSetup1 = globalActions.commitmentReceived({
-  processId: channelId,
-  signedCommitment: scenarios.signedCommitment3,
-});
+
+const sharedData = () => ({ ...advanceChannelScenarios.initialStore });
```
lalexgap: Is there a reason we need this function?

packages/wallet/src/redux/protocols/direct-funding/__tests__/scenarios.ts
```diff
@@ -73,78 +51,68 @@ const bFundingReceivedEvent = globalActions.fundingReceivedEvent({
...
   },
   waitForPostFundSetup: {
     state: states.waitForFundingAndPostFundSetup({
       ...defaultsForA,
       channelFunded: true,
-      postFundSetupReceived: false,
     }),
-    sharedData: aHasPostFund0SharedData,
-    action: postFundSetup1,
+    sharedData: { ...advanceChannelScenarios.initialStore },
```
lalexgap: Should this be `sharedData()` to be consistent?

packages/wallet/src/redux/protocols/direct-funding/__tests__/scenarios.ts
andrewgordstewart: I'm not sure. We re-assign to `sharedData` at many points in reducers, so the value of `advanceChannelScenarios.initialStore` may change throughout a test suite.

packages/wallet/src/redux/protocols/direct-funding/__tests__/scenarios.ts
andrewgordstewart:
```suggestion
    sharedData: sharedData(),
```

PR #546, opened 2019-06-22
Fix existing channel reducer

### LINKED ISSUE #545: ExistingChannelFunding Reducer should take an allocation/destination not total

lalexgap:
Fixes #545 

Previously the existing channel reducer assumed that both players would equally fund a channel. While this worked for RPS(since each player funds half of the total) but would not work in the general case.

This fixes the issue by using the allocation/destination of the app channel to figure out if the ledger channel needs to be topped up.


packages/wallet/src/redux/protocols/existing-channel-funding/reducer.ts
```diff
@@ -271,46 +267,36 @@ const waitForLedgerUpdateReducer = (
...
   }
-  const numParticipants = helpers.getNumberOfParticipants(latestCommitment);
-  const amountRequiredFromEachParticipant = bigNumberify(proposedAmount).div(numParticipants);
-
-  return !latestCommitment.allocation.every(a =>
-    bigNumberify(a).gte(amountRequiredFromEachParticipant),
-  );
+  // We assume that destination/allocation are the same length and destination contains the same addresses
+  // Otherwise we shouldn't be in this protocol at all
+  for (let i = 0; i < proposedDestination.length; i++) {
```
andrewgordstewart: This could be something like
```
return proposedDestination.some(address => {
  const existingIndex = latestCommitment.destination.indexOf(address);
  return bigNumberify(latestCommitment.allocation[existingIndex]).lt(proposedAllocation[i]);
}
``

lalexgap:
> Looks ok, but is there a test that exhibits the new behaviour? Ie. it would fail on master and would pass with this change?

I've changed the balances to be asymmetric that should catch the problem. Also added some basic top up tests.

PR #548, opened 2019-06-24
Add protocolLocator to `AdvanceChannel` state

andrewgordstewart:


packages/wallet/src/redux/protocols/advance-channel/actions.ts
```diff
@@ -18,7 +18,7 @@ export const clearedToSend: ActionConstructor<ClearedToSend> = p => {
 
 export function isAdvanceChannelAction(action: WalletAction): action is AdvanceChannelAction {
   return (
-    action.type === 'WALLET.ADVANCE_CHANNEL.COMMITMENTS_RECEIVED' ||
+    action.type === 'WALLET.COMMON.COMMITMENTS_RECEIVED' ||
```
lalexgap: Now that we're making use of the `ProtocolLocator` we can add a check for `ADVANCE_CHANNEL_PROTOCOL_LOCATOR`

packages/wallet/src/redux/protocols/advance-channel/reducer.ts
```diff
@@ -154,6 +155,7 @@ function initializeWithNewChannel(
       nextParticipant(participants, ourIndex),
       processId,
       [signResult.signedCommitment],
+      ADVANCE_CHANNEL_PROTOCOL_LOCATOR,
```
lalexgap: Shouldn't this be `protocolLocator` from state?

andrewgordstewart: From what state?

lalexgap: It should be the`protocolLocator` variable that is passed into `initialize`. That way the action that gets dispatched has the full `protocolLocator` string (IE:`DirectFunding-AdvanceChannel`) as opposed to just `AdvanceChannel.

PR #550, opened 2019-06-25
Update Consensus Update protocol to support n players

### LINKED ISSUE #549: Refactor `UpdateConsensus` to allow for n-party state channels.

lalexgap:
Updates the consensus update protocol to support any arbitrary amount of players.

Fixes #549.


packages/wallet/src/redux/protocols/advance-channel/reducer.ts
```diff
@@ -150,13 +145,12 @@ function initializeWithNewChannel(
     sharedData = registerChannelToMonitor(sharedData, processId, channelId);
 
     // Send commitments to next participant
-    const messageRelay = sendCommitmentsReceived(
-      nextParticipant(participants, ourIndex),
+    sharedData = helpers.sendCommitments(
```
andrewgordstewart: 👍

packages/wallet/src/redux/protocols/consensus-update/__tests__/reducer.test.ts
```diff
@@ -86,3 +197,23 @@ function itTransitionsTo(
     expect(result.protocolState.type).toEqual(type);
   });
 }
+
+function itSendsMessage(
```
andrewgordstewart: I think there's already a helper that does this, that's used in the advance channel tests: `expectCommitmentsSent`

packages/wallet/src/redux/protocols/consensus-update/__tests__/scenarios.ts
```diff
@@ -36,92 +53,267 @@ const ledger7 = ledgerCommitment({
...
+});
+
 // ------
 // SharedData
 // ------
-const aInitialSharedData = setChannels(EMPTY_SHARED_DATA, [
+
+const threePlayerInitialSharedData = (ourIndex: ThreePartyPlayerIndex) => {
+  const address =
+    ourIndex === ThreePartyPlayerIndex.A
```
andrewgordstewart: Why not have a lookup table
```
{
  [ThreePartyPlayerIndex.A]: asAddress,
  ...
}
```

This lookup table would be reused a bunch of times in the next 80 lines of code or so.

packages/wallet/src/redux/protocols/consensus-update/__tests__/reducer.test.ts
```diff
@@ -3,78 +3,189 @@ import { initialize, consensusUpdateReducer } from '../reducer';
...
+      } = scenario.initialize;
+      const result = initialize(
+        processId,
+        channelId,
+        proposedAllocation,
+        proposedDestination,
+        sharedData,
+      );
+
+      itTransitionsTo(result, 'ConsensusUpdate.WaitForUpdate');
```
andrewgordstewart: Add `itSendsNoMessage`.

packages/wallet/src/redux/protocols/consensus-update/__tests__/reducer.test.ts
```diff
@@ -3,78 +3,189 @@ import { initialize, consensusUpdateReducer } from '../reducer';
 import * as states from '../states';
```
andrewgordstewart: It'd be nice to add `itSendsNoMessage` checks in any example that doesn't have an `itSendsMessage` check.

packages/wallet/src/redux/protocols/consensus-update/reducer.ts
```diff
@@ -56,75 +64,102 @@ export const consensusUpdateReducer = (
...
     console.warn(`Consensus update reducer was called with terminal state ${protocolState.type}`);
     return { protocolState, sharedData };
   }
+  const { channelId, processId } = protocolState;
+
+  try {
+    const { turnNum } = getExistingChannel(sharedData, channelId);
+    sharedData = helpers.checkCommitments(sharedData, turnNum, action.signedCommitments);
+  } catch (err) {
+    return { protocolState: states.failure({ reason: 'UnableToValidate' }), sharedData };
```
andrewgordstewart: Should we add the error to the reason to help debugging?

andrewgordstewart:
```suggestion
    return { protocolState: states.failure({ reason: `UnableToValidate: ${err.message}` }), sharedData };
```

packages/wallet/src/redux/protocols/consensus-update/reducer.ts
```diff
@@ -56,75 +64,102 @@ export const consensusUpdateReducer = (
...
   }
 
-  return theirAddress(channel);
-}
-function getLatestCommitment(sharedData: SharedData, channelId: string) {
-  const channel = getChannel(sharedData, channelId);
-  if (!channel) {
-    throw new Error(`Could not find channel for id ${channelId}`);
+  for (let i = 0; i < expectedAllocation.length; i++) {
+    if (allocation[i] !== expectedAllocation[i] || destination[i] !== expectedDestination[i]) {
```
andrewgordstewart: This should probably be `eqHex(allocation[i], expectedAllocation[i])`.

As is, I think the current implementation is equivalent to
```
return allocation.toString() === expectedAllocation.toString() && destination.toString() === expectedDestination.toString()
```
but I don't think that's what we want to check.

andrewgordstewart: We can probably add an `eqHexArray` utility, and should use those.

packages/wallet/src/redux/protocols/reducer-helpers.ts
```diff
@@ -88,6 +89,45 @@ export function sendOpponentConcluded(sharedData: SharedData): SharedData {
...
+  // exceed the current turnNum.
+
+  commitments
+    .filter(signedCommitment => signedCommitment.commitment.turnNum > turnNum)
+    .map(signedCommitment => {
+      const result = checkAndStore(sharedData, signedCommitment);
+      if (result.isSuccess) {
+        sharedData = result.store;
+      } else {
+        throw new Error('Unable to validate commitment');
```
andrewgordstewart: Should we rewrite this to return `{ isSuccess: false, reason: `Unable to validate commitment: ${result.reason}` }` instead?

packages/wallet/src/redux/channel-store/channel-state/states.ts
```diff
@@ -72,7 +72,8 @@ export function pushCommitment(
 export function ourTurn(state: ChannelState) {
   const { turnNum, participants, ourIndex } = state;
   const numParticipants = participants.length;
-  return turnNum % numParticipants !== ourIndex;
+
+  return (turnNum + 1) % numParticipants === ourIndex;
```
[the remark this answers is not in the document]
lalexgap: It was correct for 2 players not n players.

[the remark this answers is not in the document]
andrewgordstewart: Oh, I missed replacing `!==` with `===`

packages/wallet/src/domain/consensus-app/index.ts
```diff
@@ -10,13 +10,28 @@ import {
...
 } from 'fmg-nitro-adjudicator/lib/consensus-app';
 import { Commitment } from 'fmg-core';
 import { CommitmentType } from '../commitments';
 /////////////
 // Helpers //
 /////////////
 
+export function consensusHasBeenReached(commitment: Commitment): boolean {
+  const consensusCommitment = fromCoreCommitment(commitment);
+  return consensusCommitment.appAttributes.updateType === UpdateType.Consensus;
```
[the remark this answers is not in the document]
lalexgap: I don't think `fmg-nitro-adjudicator` has been published with those changes. The latest version is `0.1.2` which looks like it's missing your most recent changes: https://github.com/magmo/force-move-protocol/compare/fmg-nitro-adjudicator@0.1.2...master

packages/wallet/src/domain/commitments/__tests__/index.ts
```diff
@@ -116,6 +131,51 @@ export const ledgerChannel = {
 };
 export const ledgerId = channelID(ledgerChannel);
 
+export const threeWayLedgerChannel = {
```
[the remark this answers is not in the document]
lalexgap: I'm following the newer pattern of using the test helper functions that construct `channel`s for you as opposed to using static test data.

I think the big benefit is that a test can create the type of `channel` it needs as opposed to relying on existing test data working for its use case. This is a less of an issue with joint channels since they're used on in a couple of places but I still think it makes sense to follow the same pattern.

packages/wallet/src/domain/commitments/__tests__/index.ts
```diff
@@ -19,16 +24,17 @@ export const channelId = channelID(channel);
 function typeAndCount(
   turnNum: number,
   isFinal: boolean,
+  numParticipants = 2,
 ): { commitmentCount: number; commitmentType: CommitmentType } {
   let commitmentCount;
   let commitmentType;
   if (isFinal) {
-    commitmentCount = turnNum % 2;
+    commitmentCount = turnNum % numParticipants;
```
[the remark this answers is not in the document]
lalexgap: Yeah, you're right, this won't always work correctly.

packages/wallet/src/domain/commitments/__tests__/index.ts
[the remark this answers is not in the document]
andrewgordstewart: I agree. I wonder if we should
A. Migrate towards that pattern
B. Consolidate the test data into one place

I would have used this approach, except I found the static test data in another file.

packages/wallet/src/domain/consensus-app/index.ts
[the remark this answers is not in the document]
andrewgordstewart: Ok, should we leave it as is, and fix it after publishing the changes?

packages/wallet/src/redux/protocols/reducer-helpers.ts
lalexgap: This was a function you wrote ;)
Are you suggesting the return type be `{sharedData:SharedData, isSuccess:boolean, reason?:string}`?

andrewgordstewart: Maybe? I'm not really up to speed on our approach to handling errors.

packages/wallet/src/domain/consensus-app/index.ts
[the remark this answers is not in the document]
lalexgap: That makes sense to me!

packages/wallet/src/redux/protocols/reducer-helpers.ts
lalexgap: Unfortunately I don't think we have a unified approach at the moment 😢

PR #551, opened 2019-06-25
Use redux-saga to dispatch multiple actions

geoknee:
This enables the user to 
a) terminate the `Dispute-` process
b) start a new `Defund-` process 

with one click (but two actions). Previously, one action was consumed by two reducers. 

An early version of this PR used `redux-thunk` to get the job done, but I have since replaced that with a `redux-saga` solution, which moreover allows for an arbitrary array of actions to be dispatched.

### DIFF
--- a/packages/wallet/src/redux/actions.ts
+++ b/packages/wallet/src/redux/actions.ts
@@ -28,6 +28,11 @@ export const isTransactionAction = isTA;
 // -------
 // Actions
 // -------
+
+export interface MultipleWalletActions {
+  type: 'MULTIPLE_ACTIONS';
+  actions: WalletAction[];
+}
 export interface LoggedIn {
   type: 'WALLET.LOGGED_IN';
   uid: string;
@@ -116,6 +121,11 @@ export interface ChallengeExpiredEvent {
 // -------
 // Constructors
 // -------
+
+export const MultipleWalletActions: ActionConstructor<MultipleWalletActions> = p => ({
+  ...p,
+  type: 'MULTIPLE_ACTIONS',
+});
 
 export const loggedIn: ActionConstructor<LoggedIn> = p => ({ ...p, type: 'WALLET.LOGGED_IN' });
 
--- a/packages/wallet/src/redux/actions.ts
+++ b/packages/wallet/src/redux/actions.ts
@@ -122,7 +122,7 @@ export interface ChallengeExpiredEvent {
 // Constructors
 // -------
 
-export const MultipleWalletActions: ActionConstructor<MultipleWalletActions> = p => ({
+export const multipleWalletActions: ActionConstructor<MultipleWalletActions> = p => ({
   ...p,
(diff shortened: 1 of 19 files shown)


geoknee:
Following conversations with @andrewgordstewart and @lalexgap, I'm going to try and replicate this with `redux-saga`, so that we don't use more middlewares than we need to. 

The new saga will be something along the lines of `dispatchMultipleActions`: it will watch for a new action type `ActionArray`, and simply dispatch all of the actions in the array. That way, it will be clear which actions you wish to dispatch from the container.

packages/wallet/src/redux/actions.ts
```diff
@@ -117,6 +122,11 @@ export interface ChallengeExpiredEvent {
 // Constructors
 // -------
 
+export const multipleWalletActions: ActionConstructor<MultipleWalletActions> = p => ({
```
kerzhner: Even though you are mimicking neighboring code, I am going to comment anyways :). It would be great to use more descriptive parameter names than `p`.

packages/wallet/src/redux/protocols/dispute/challenger/actions.ts
```diff
@@ -23,6 +22,10 @@ export interface ChallengeDenied {
   processId: string;
 }
 
+export interface ExitChallenge {
+  type: 'WALLET.DISPUTE.CHALLENGER.EXIT_CHALLENGE';
+  processId: string;
+}
```
kerzhner:
```suggestion
}

```

packages/wallet/src/redux/sagas/multiple-action-dispatcher.ts
```diff
@@ -0,0 +1,9 @@
+import { MultipleWalletActions } from '../actions';
+import { take, put } from 'redux-saga/effects';
+
+export function* multipleActionDispatcher() {
+  while (true) {
+    const multipleWalletActions: MultipleWalletActions = yield take('WALLET.MULTIPLE_ACTIONS');
+    yield multipleWalletActions.actions.map(action => put(action));
```
kerzhner: Does this work? Should `yield` apply to the `map` function or the `put` function?

packages/wallet/src/redux/actions.ts
geoknee: Agreed! This pattern is prolific, but I think we could clear it up in a separate PR. Something like `props` would probably be an improvement?

packages/wallet/src/redux/sagas/multiple-action-dispatcher.ts
geoknee: It works: first thing I tries was to `yield put` inside the loop, but that *doesn't* work.

kerzhner: I think I understand what is going now. The saga is a generator that `yield`s an array. Then, the iterator (in this case the saga framework), processes the contents of the array.

PR #553, opened 2019-06-26
Improve commitment test helpers

andrewgordstewart:
See commit messages and comments.

### DIFF
--- a/packages/wallet/src/redux/__tests__/helpers.ts
+++ b/packages/wallet/src/redux/__tests__/helpers.ts
@@ -84,7 +84,10 @@ type PartialCommitments = Array<{ commitment: Partial<Commitment>; signature?: s
 
 function transformCommitmentToMatcher(sc: { commitment: Partial<Commitment>; signature?: string }) {
   if (sc.signature) {
-    return expect.objectContaining({ signature: sc.signature });
+    return expect.objectContaining({
+      commitment: expect.objectContaining(sc.commitment),
+      signature: sc.signature,
+    });
   } else {
     return expect.objectContaining({ commitment: expect.objectContaining(sc.commitment) });
   }
--- a/packages/wallet/src/redux/__tests__/helpers.ts
+++ b/packages/wallet/src/redux/__tests__/helpers.ts
@@ -126,6 +126,7 @@ export const expectThisCommitmentSent = (
       // To help with debugging, you can change the idx variable when running tests to 'search'
       // for the correct commitment
 
+      console.warn(`Message not found: inspecting mismatched message in position ${idx}`);
       expect(messageOutbox[idx]).toMatchObject({
         messagePayload: {
           type,
@@ -170,6 +171,7 @@ export const expectTheseCommitmentsSent = (
 
       // To help with debugging, you can change the idx variable when running tests to 'search'
       // for the correct commitment
+      console.warn(`Message not found: inspecting mismatched message in position ${idx}`);
       expect(messageOutbox[idx]).toMatchObject({
         messagePayload: {
           type,
--- a/packages/wallet/src/redux/__tests__/helpers.ts
+++ b/packages/wallet/src/redux/__tests__/helpers.ts
@@ -82,10 +82,7 @@ export const expectThisMessage = (state: SideEffectState, messageType: string) =
 
 type PartialCommitments = Array<{ commitment: Partial<Commitment>; signature?: string }>;
 
-export function transformPartialCommitmentToObjectContainint(sc: {
-  commitment: Partial<Commitment>;
-  signature?: string;
-}) {
+function transformCommitmentToMatcher(sc: { commitment: Partial<Commitment>; signature?: string }) {
   if (sc.signature) {
     return expect.objectContaining({ signature: sc.signature });
   } else {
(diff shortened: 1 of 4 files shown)


packages/wallet/src/redux/__tests__/helpers.ts
```diff
@@ -79,51 +79,114 @@ export const expectThisMessage = (state: SideEffectState, messageType: string) =
...
-  expectSideEffect('messageOutbox', state, item => {
-    expect(item.messagePayload.type).toEqual(messageType);
-    expect(item.messagePayload.signedCommitment.commitment).toMatchObject(c);
-  });
-};
 
-export const expectThisMessageAndTheseCommitmentsSent = (
+type PartialCommitments = Array<{ commitment: Partial<Commitment>; signature?: string }>;
+
+export function transformPartialCommitmentToObjectContainint(sc: {
```
geoknee: Typo here Containint -> Containing (presumably).

Could this function be better named?. Unless I misunderstand, the input type is not a `PartialCommitment` (it is an unnamed type which has a property with type `PartialCommitment`), and it does not return an `Object` (it returns a jest matcher).

andrewgordstewart: a Jest matcher is an object!

andrewgordstewart: I think it's too wordy to say `transformFromMaybeSignedPartialCommitmentToJestMatcher`. I would be more inclined to say `transformCommitmentToMatcher`. There's no questioning what kind of object it takes -- it's right there in the signature!

packages/wallet/src/redux/__tests__/helpers.ts
```diff
@@ -74,51 +79,118 @@ export const expectThisMessage = (state: SideEffectState, messageType: string) =
...
-    expect(item.messagePayload.signedCommitment.commitment).toMatchObject(c);
-  });
-};
 
-export const expectThisMessageAndTheseCommitmentsSent = (
+type PartialCommitments = Array<{ commitment: Partial<Commitment>; signature?: string }>;
+
+function transformCommitmentToMatcher(sc: { commitment: Partial<Commitment>; signature?: string }) {
+  if (sc.signature) {
+    return expect.objectContaining({ signature: sc.signature });
```
kerzhner: Why does the expected object not contain a commitment property?

packages/wallet/src/redux/__tests__/helpers.ts
```diff
@@ -74,51 +79,118 @@ export const expectThisMessage = (state: SideEffectState, messageType: string) =
...
+      // We try to help the developer by using expect(foo).toMatchObject(bar)
+      // The errors are much more useful in this case, but will be deceiving in the case when
+      // multiple messages are queued.
+
+      // To help with debugging, you can change the idx variable when running tests to 'search'
+      // for the correct commitment
 
-export const expectThisCommitmentSent = (state: SideEffectState, c: Partial<Commitment>) => {
-  expectThisMessageAndCommitmentSent(state, c, 'WALLET.COMMON.COMMITMENT_RECEIVED');
+      expect(messageOutbox[idx]).toMatchObject({
```
kerzhner: What about a `console.warn` message that points the developer to this code. Something like: 'Make sure to understand the matcher before diving into debugging'?

packages/wallet/src/redux/__tests__/helpers.ts
```diff
@@ -74,51 +79,118 @@ export const expectThisMessage = (state: SideEffectState, messageType: string) =
...
+        },
+      });
+    } else {
+      throw err;
+    }
+  }
 };
-type PartialCommitments = Array<{ commitment: Partial<Commitment>; signature?: string }>;
 
 export const expectTheseCommitmentsSent = (
```
kerzhner: Should this use `expectThisCommitmentSent`?

packages/wallet/src/redux/__tests__/helpers.ts
```diff
@@ -74,51 +79,118 @@ export const expectThisMessage = (state: SideEffectState, messageType: string) =
...
+        },
+      });
+    } else {
+      throw err;
+    }
+  }
 };
 
+function getOutboxState(state: SideEffectState, outboxBranch: 'messageOutbox'): MessageOutbox {
+  if ('sideEffects' in state && state.sideEffects && state.sideEffects[outboxBranch]) {
```
kerzhner: Why is `'sideEffects' in state` needed?

packages/wallet/src/redux/__tests__/helpers.ts
andrewgordstewart: Good catch!

packages/wallet/src/redux/__tests__/helpers.ts
andrewgordstewart: `SideEffectState` is a union type, where not all members have the `sideEffect` property. So, without, the compiler is unaware that `state` can have a `sideEffect` property.

packages/wallet/src/redux/__tests__/helpers.ts
andrewgordstewart: No, because `expectThisCommitmentSent` is about sending `COMMITMENT_RECEIVED`, which only works for two-party channels.

In an n-party channel, we are currently sending the whole round of commitments, as this is the only thing that allows "forward-only" communication to work. This is a different action: `COMMITMENTS_RECEIVED`

kerzhner: Gotcha. It's unfortunate that most of the functionality between the `expectThisCommitmentSent` and `expectTheseCommitmentsSent` is copy-pasted. But this is test code after all...

andrewgordstewart: Yes, I thought about trying to dry it out a bit, but deemed it not worth the effort.

PR #562, opened 2019-06-27
Validate Transitions

### LINKED ISSUE #561: Validate transitions
A production wallet needs to guarantee that it can launch a force-move/refutation/conclusion. This requires a valid round (or at least a valid pair) of commitments. That is, each successive pair of commitments forms a valid transition.

Currently the wallet does not guarantee this. The primary reason is that a reducer has to be a pure, sync function. Thus, it can't call a smart contract function to validate transitions according to the "ground-truth" rules.

lalexgap:
Calls `validateTransition` from the `message-listener` whenever receiving a `commitment`.

Fixes #561.

### DIFF
--- a/packages/wallet/src/redux/sagas/message-listener.ts
+++ b/packages/wallet/src/redux/sagas/message-listener.ts
@@ -1,4 +1,4 @@
-import { take, put } from 'redux-saga/effects';
+import { take, put, select } from 'redux-saga/effects';
 import * as incoming from 'magmo-wallet-client/lib/wallet-instructions';
 
 import * as actions from '../actions';
@@ -6,9 +6,11 @@ import { eventChannel } from 'redux-saga';
 import * as application from '../protocols/application/reducer';
 import { isRelayableAction, WalletProtocol } from '../../communication';
 import { responseProvided } from '../protocols/dispute/responder/actions';
-import { getChannelId } from '../../domain';
+import { getChannelId, Commitment, SignedCommitment } from '../../domain';
 import { concluded } from '../protocols/application/actions';
-
+import * as selectors from '../selectors';
+import * as contractUtils from '../../utils/contract-utils';
+import { appAttributesFromBytes } from 'fmg-nitro-adjudicator';
 export function* messageListener() {
   const postMessageEventChannel = eventChannel(emitter => {
     window.addEventListener('message', (event: MessageEvent) => {
@@ -56,6 +58,8 @@ export function* messageListener() {
             actions.protocol.initializeChannel({ channelId: getChannelId(action.commitment) }),
           );
         }
+        yield validateAgainstLatestCommitment(action.commitment);
+
         yield put(
           actions.application.ownCommitmentReceived({
             processId: application.APPLICATION_PROCESS_ID,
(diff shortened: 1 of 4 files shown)


packages/wallet/src/redux/sagas/message-listener.ts
```diff
@@ -94,6 +108,59 @@ export function* messageListener() {
...
+  // If we're receiving the first commitment there's nothing stored to validate against
+  if (signedCommitments[0].commitment.turnNum > 0) {
+    const latestStoredCommitment: Commitment = yield select(
+      selectors.getLastCommitmentForChannel,
+      channelId,
+    );
+    const newCommitments = signedCommitments.filter(
+      signedCommitment => signedCommitment.commitment.turnNum > latestStoredCommitment.turnNum,
+    );
+    if (newCommitments.length > 0) {
```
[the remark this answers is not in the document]
andrewgordstewart: Considering that, we can just do
```
signedCommitments.filter(
  signedCommitment => signedCommitment.commitment.turnNum > latestStoredCommitment.turnNum,
).reduce( [above code] )
```

packages/wallet/src/redux/sagas/message-listener.ts
```diff
@@ -94,6 +108,59 @@ export function* messageListener() {
   }
 }
 
+function* validateTransitionForCommitments(signedCommitments: SignedCommitment[]) {
+  const channelId = getChannelId(signedCommitments[0].commitment);
+  // If we're receiving the first commitment there's nothing stored to validate against
+  if (signedCommitments[0].commitment.turnNum > 0) {
```
andrewgordstewart: I think we would want to ask directly if we have something stored to validate against. If B sends [{turnNum: 0}, {turnNum: 1}]` to A, then we wouldn't enter this block.

packages/wallet/src/redux/sagas/message-listener.ts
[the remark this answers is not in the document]
lalexgap: `yield` is not allowed inside a reducer function.

packages/wallet/src/redux/sagas/message-listener.ts
lalexgap: Updated

packages/wallet/src/redux/sagas/message-listener.ts
```diff
@@ -94,6 +108,60 @@ export function* messageListener() {
   }
 }
 
+function* validateTransitionForCommitments(signedCommitments: SignedCommitment[]) {
+  const channelId = getChannelId(signedCommitments[0].commitment);
+
+  const storedCommitmentExists = yield select(selectors.doesACommitmentExistForChannel, channelId);
```
andrewgordstewart: An alternative here is to rewrite `getLastCommitmentForChannel` to throw when the channel isn't stored, catch that error, and do nothing.

(Or have it return `undefined`, and if it's undefined, do nothing)

lalexgap: I think I'd prefer the extra `selector` so I'm going to keep it that way for now.

PR #564, opened 2019-06-28
Remove http hub special case

### LINKED ISSUE #557: Move away from http for client to server communication

kerzhner:
Fixes #557 and #569.

- A permanent entry is added to the challenges (not to be confused with force move protocol challenges) in the rock-paper-scissors-dev realtime database for the static hub address. This is problematic if multiple developers are running a hub locally. To be addressed in a future PR.
- `yarn firebase-relay:start:dev` runs a service locally that monitors the realtime database and relays messages to the server. In staging and production environments, a cloud function replaces this service.
- Server responses are written to the realtime database.
- `v1` server routes are removed. `v2` server routes are renamed to `v1`.


packages/server/src/message/firebase-relay.ts
```diff
@@ -0,0 +1,64 @@
...
+  hubRef.on('child_added', snapshot => {
+    const key = snapshot.key;
+    const value = snapshot.val();
+    postData({ ...value.payload, queue: value.queue });
+    hubRef.child(key).remove();
+  });
+}
+
+export function send(to, payload) {
+  const fbPayload = { payload, queue: 'WALLET' };
```
kerzhner: Not sure that we should be hardcoding the queue here.

packages/server/.vscode/launch.json
```diff
@@ -54,6 +54,19 @@
...
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Launch Message Relay",
+      "cwd": "${workspaceRoot}",
+      "protocol": "inspector",
+      "internalConsoleOptions": "neverOpen",
+      "runtimeArgs": ["-r", "ts-node/register"],
+      "args": ["${workspaceFolder}/src/message/firebase-relay.ts"],
```
andrewgordstewart: What's the benefit of this being a VSCode debug command?

packages/server/src/app/app.ts
```diff
@@ -9,6 +10,17 @@ import { channelRoutes } from './routes/v2/channels';
 
 const app = new Koa();
 
+app.use(async (ctx, next) => {
+  await next();
```
andrewgordstewart: Can this be extracted to a function in its own file? It would help keep `app.ts` minimal.

packages/server/src/app/routes/v2/channels.ts
```diff
@@ -10,17 +11,35 @@ export const BASE_URL = `/api/v2/channels`;
...
+    sendViaFirebase(response);
+    return response;
   } else {
     const action = ctx.request.body;
 
     if (await isNewProcessAction(action)) {
-      return await handleNewProcessAction(ctx);
+      const response = await handleNewProcessAction(ctx);
+      sendViaFirebase(response);
+      return response;
```
andrewgordstewart: If we are sending commitments via firebase, do we need to return the commitments in the http request? Should we instead return something that's useful to the firebase function?

packages/server/src/message/firebase-relay.ts
```diff
@@ -25,11 +25,22 @@ const config = {
   messagingSenderId: '913007764573',
 };
 
-const firebase = firebaseApp.initializeApp(config);
-const messagesRef = firebase.database().ref('messages');
+let firebase;
```
andrewgordstewart: What's the purpose of dynamically instantiating this?

packages/server/.vscode/launch.json
kerzhner: `firebase-relay is run as a separate process. For debugging, it's nice to start the process through vscode for breakpoint setting, etc.

packages/server/src/message/firebase-relay.ts
kerzhner: `initializeApp` should only be called once for a configuration. So I am using the singleton pattern.

packages/server/src/app/routes/v2/channels.ts
kerzhner: Firebase relay removes the message from the realtime database after forwarding the message to the hub. So any information from the hub that signals whether to remove the message is useful. No need to reply with a commitment. A response code is sufficient.

packages/server/src/app/app.ts
kerzhner: Good point. This was resolved in a later commit

PR #572, opened 2019-07-01
Clean up

geoknee:


packages/rps/src/redux/message-service/saga.ts
```diff
@@ -210,7 +210,7 @@ function createWalletEventChannel(walletEventTypes: Wallet.WalletEventType[]) {
     return () => {
       listener.unSubscribeAll();
     };
-  });
+  }, buffers.expanding(10));
```
geoknee: According to the `redux-saga` [docs](https://redux-saga.js.org/docs/advanced/Channels.html), `eventChannel`s do not buffer messages by default (unlike `actionChannel`s). If you `take` from an unbuffered channel, and then do a blocking `yield call`, you can miss the next message. 

In the past the workaround was to use a non-blocking `yield fork` in place of the `yield call`, but the downside to this is that the messages are processed concurrently. This solution is superior as it will maintain the order of the messages (i.e they are processed serially). 

It is important to have this functionality work, because we may need to send multiple messages from the wallet to be relayed by the app. One can either 
i) use `queueMessage(queueMessage(...))` _or_
ii) relay a batch of messages instead. (see #555 ). 

Since both methods are in use in the wallet, they need to be supported by this saga in the app.

packages/rps/src/redux/message-service/saga.ts
```diff
@@ -48,7 +48,7 @@ interface WalletMessage {
 type Message = AppMessage | WalletMessage;
 
 export function* sendWalletMessageSaga() {
-  const sendMessageChannel = createWalletEventChannel([Wallet.MESSAGE_RELAY_REQUESTED]);
+  const sendMessageChannel = yield call(createWalletEventChannel, [Wallet.MESSAGE_RELAY_REQUESTED]);
```
kerzhner: Any specific reason to convert this function invocation to the `yield call`? Typically, `yield call` is employed to make side effects unit-testable...

packages/rps/src/redux/message-service/saga.ts
kerzhner: I wonder if it is safer to use `buffers.fixed(limit)`. The benefit of a fixed limit is that, if an event channel has no subscribers, we will see an error after a certain number of events are buffered. With an expanding buffer, I am guessing that the browser will slow down and/or hang after a while.

packages/rps/src/redux/message-service/saga.ts
geoknee: I was following the tutorial in the redux-saga docs, and their examples do it this way. I had also read that not `yield`-ing effects inside a `while` loop can cause the app to hang (that's not an issue here but that may have influenced me to use the `yield call` here). 

 Looking in to it further however, I think you're right that it only seems to serve to make testing easier. Reverting this for now.

packages/rps/src/redux/message-service/saga.ts
geoknee: Seems fair, have gone for the fixed limit of 10. As per our discussions, we may be moving towards a single 'batched' message for communication, which would then make this buffering issue somewhat academic 👍

PR #577, opened 2019-07-02
Storybook overhaul

geoknee:
All protocols should feature test scenarios, and we would like to use storybook to preview each of the UI screens corresponding to the progression of states in each scenario.

Previously, the approach was to recursively nest a given protocol state inside its parents until hitting a top level protocol, and then have this (rather bloated) state fed into the top level wallet container. 

This PR implements a more scalable approach: when adding a scenario to storybook, simply pass in the relevant protocol container as an additional parameter. The passed-in container handles the wiring between states and their components,  and a new render method handle wraps the resulting component in a modal in the same way that the wallet container would.

This PR has:

Downsides: 
- the styling of the wallet container has been somewhat duplicated in the new render function
- storybook will show the components, but depending on whether they are wired up correctly to a top level container they may not show in the app

Upsides:
- No complicated nesting to top level is required, and less mocking is required
- storybook behaves more like a unit test and less like an integration test


packages/wallet/src/__stories__/index.tsx
```diff
@@ -16,27 +17,28 @@ const walletStateRender = state => () => {
...
+  return (
+    <Provider store={fakeStore(protocolState)}>
+      <Modal
+        isOpen={true}
+        className={'wallet-content-center'}
+        overlayClassName={'wallet-overlay-center'}
+        ariaHideApp={false}
+      >
+        <StatusBarLayout>
+          <Container state={protocolState} />
```
kerzhner: Does this work for containers that are not "top-level"? In other words, what if a container is passed that the wallet displays inside another container?

geoknee: That's not a problem currently, since all the containers do as a whole is to consume a top level protocol state, drill down to the lowest level protocol state that is nested, and then return the appropriate component. In the wild, a non-top-level container is going to return this component to its parent container, and on to its parent (etc). Here we just shortcut that process and get the component directly from its neatest container. It can then be wrapped in a bit more frontend code and rendered.

PR #578, opened 2019-07-02
Use shared components

geoknee:
This PR changes our containers to reuse our react components as much as possible, and deletes most of the 'custom' protocol components that exist in the wallet.

We have some pretty good general purpose components such as `Approve-X`. Reusing those will make consistent styling much easier in future. 

I have preserved some of the components that are more sophisticated than the shared components: for example,`funding-step.tsx` and a couple of others


packages/wallet/src/redux/protocols/concluding/instigator/container.tsx
```diff
@@ -44,25 +42,38 @@ class ConcludingContainer extends PureComponent<Props> {
...
+            approvalAction={() => approve({ processId })}
           />
         );
       case 'ConcludingInstigator.WaitForLedgerUpdate':
         return <ConsensusUpdate state={state.consensusUpdateState} />;
       case 'ConcludingInstigator.WaitForOpponentSelection':
-        return <WaitForOpponentDecision />;
+        return (
+          <WaitForOtherPlayer
+            name={'decision about defunding the ledger channel'}
```
kerzhner: Even though `WaitForTheOtherPlayer` was not created as part of this PR, I am leaving this comment. Is there a better property description than `name`? `action-descriptor`?

geoknee: :+1:

PR #591, opened 2019-07-05
Remove old test scenarios

### LINKED ISSUE #590: Remove test-scenarios file
Test data is generated in two places:
https://github.com/magmo/apps/blob/master/packages/wallet/src/domain/commitments/__tests__/index.ts
and
https://github.com/magmo/apps/blob/master/packages/wallet/src/redux/__tests__/test-scenarios.ts

The first has the benefit of being able to specify the commitment data, with sensible defaults.

We should remove the second file, and adjust protocol scenarios that depend on it.

lalexgap:
Removes the older `test-scenarios` so we only get our test data from one place. Fixes #590


packages/wallet/src/utils/__tests__/commitment-utils.test.ts
```diff
@@ -1,10 +1,9 @@
-import { appCommitment } from '../../domain/commitments/__tests__';
+import { appCommitment, asAddress, asPrivateKey } from '../../domain/commitments/__tests__';
 import { channelFromCommitments } from '../../redux/channel-store/channel-state/__tests__';
-import * as channelScenarios from '../../redux/__tests__/test-scenarios';
+
```
andrewgordstewart:
```suggestion
```

PR #592, opened 2019-07-08
Storage and Locking PoC

lalexgap:


packages/wallet/src/redux/protocols/existing-ledger-funding/reducer.ts
```diff
@@ -29,70 +38,33 @@ export const initialize = (
...
-    const messageRelay = sendCommitmentReceived(
-      theirAddress(ledgerChannel),
-      processId,
-      signResult.signedCommitment.commitment,
-      signResult.signedCommitment.signature,
-      EXISTING_LEDGER_FUNDING_PROTOCOL_LOCATOR,
-    );
-    sharedData = queueMessage(sharedData, messageRelay);
+  sharedData = registerChannelToMonitor(sharedData, processId, ledgerId);
+  sharedData = queueLockRequest(sharedData, lockChannelRequest({ channelId: ledgerId, processId }));
```
lalexgap: Protocols can now request a lock on a channel. Currently they will wait until a lock is acquired but we can change the behavior so that a protocol gets a `CHANNEL_BUSY` action and we can handle that case appropriately.

packages/wallet/src/redux/store.ts
```diff
@@ -1,16 +1,31 @@
...
 const composeEnhancers = (window as any).__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
-const enhancers = composeEnhancers(applyMiddleware(sagaMiddleware));
+const enhancers = composeEnhancers(applyMiddleware(sagaMiddleware, storageMiddleware));
+const reducerWithStorage = storage.reducer(walletReducer);
 
-const store = createStore(walletReducer, enhancers);
+const store = createStore(reducerWithStorage, enhancers);
+const load = storage.createLoader(storageEngine);
+// TODO: Catch
+load(store).then(newState => console.log('Loaded state:', newState));
```
lalexgap: This loads the redux store from `localStorage` when the redux store is initialized. Currently we only load `channelStore`,`privateKey` and `address`

packages/rps/src/redux/message-service/saga.ts
```diff
@@ -49,20 +49,18 @@ type Message = AppMessage | WalletMessage;
...
-    const { messagePayload, to } = messageRelayRequest;
-    const messageToSend: WalletMessage = { payload: messagePayload, queue: Queue.WALLET };
-
-    yield fork(
-      reduxSagaFirebase.database.create,
-      `/messages/${to.toLowerCase()}`,
-      sanitizeMessageForFirebase(messageToSend),
-    );
-  }
+  yield takeEvery(sendMessageChannel, sendMessageSaga);
```
lalexgap: This change was required to handle the case where two `commitments_received` were queued up in the outbox.

packages/wallet/src/redux/sagas/channel-updater.ts
```diff
@@ -0,0 +1,20 @@
...
+import { select, call, put } from 'redux-saga/effects';
+import {
+  getAdjudicatorHoldings,
+  getProvider,
+  getAdjudicatorOutcome,
+} from '../../utils/contract-utils';
+import { channelUpdate } from '../actions';
+import { BigNumber, bigNumberify } from 'ethers/utils';
+
+export function* channelUpdater() {
```
lalexgap: This saga is responsible for updating the `adjudicatorState` for channels after they're loaded fro storage. Since we only store the `channelStore` in localStorage we need to update the `adjudicatorStore` with information on the funding level/ whether a channel is finalized from the blockchain.

packages/wallet/src/redux/store.ts
```diff
@@ -1,16 +1,31 @@
...
+import filter from 'redux-storage-decorator-filter';
+import createEngine from 'redux-storage-engine-localstorage';
 
+const storageEngine = filter(createEngine('magmo-wallet'), [
+  'whitelisted-key',
+  ['address'],
+  ['privateKey'],
+  ['channelStore'],
+]);
+const storageMiddleware = storage.createMiddleware(storageEngine, [
```
lalexgap: LocalStorage gets updated every time an action is dispatched. We exclude the `CHANNEL_UPDATE` action since that action should only be dispatched in response to loading from storage. Ideally if we're only storing `channelStore` we should only whitelist actions that update `channelStore`

packages/wallet/src/redux/sagas/channel-locker.ts
```diff
@@ -0,0 +1,25 @@
+import { LockChannelRequest, channelLocked, channelUnlocked } from '../actions';
+import { take, takeEvery, put, call } from 'redux-saga/effects';
+
+export function* channelLocker() {
+  yield takeEvery('WALLET.LOCKING.LOCK_CHANNEL_REQUEST', lockChannel);
+}
+
+function* lockChannel(lockRequest: LockChannelRequest) {
```
lalexgap: This is the saga responsible for locking a channel with `web-locks`. Right now it requires that any protocol that wants to (possibly) modify a channel dispatches a `LOCK_CHANNEL_REQUEST` and waits until it receives a `CHANNEL_LOCKED` action. Currently not implemented is proper handling of when a channel is already locked.

packages/wallet/src/redux/sagas/channel-locker.ts
```diff
@@ -0,0 +1,25 @@
...
+  let resolveLock;
+  const lockPromise = new Promise(resolve => (resolveLock = resolve));
+  (navigator as any).locks.request(
+    lockRequest.channelId,
+    { ifAvailable: true },
```
lalexgap: `ifAvailable:true` means that attempting to acquire a lock on something that is already locked will return right away instead of blocking. We can use this to dispatch a `CHANNEL_BUSY` action so we can handle it in the protocol and show something to the user.

packages/wallet/src/redux/protocols/channel-sync/readme.md
```diff
@@ -0,0 +1,26 @@
+# Channel Sync Protocol
```
lalexgap: The purpose of this protocol is to handle the case where one player is a `commitment` or two behind another. This could happen if one player closes their browser while another player is sending a `commitment` to them.  The protocol simply has each player exchanging the `commitment`s they have and validating `commitments` from other players. Once the protocol is complete each player should be at the same `commitment` level.

packages/wallet/src/redux/protocols/existing-ledger-funding/reducer.ts
```diff
@@ -105,10 +77,107 @@ export const existingLedgerFundingReducer = (
       return waitForPostFundSetupReducer(protocolState, sharedData, action);
     case 'ExistingLedgerFunding.WaitForLedgerTopUp':
       return waitForLedgerTopUpReducer(protocolState, sharedData, action);
+    case 'ExistingLedgerFunding.WaitForChannelSync':
+      return waitForChannelSyncReducer(protocolState, sharedData, action);
+    case 'ExistingLedgerFunding.WaitForChannelLock':
+      return waitForChannelLockReducer(protocolState, sharedData, action);
```
lalexgap: With channel locking protocols now need to wait until they can acquire a lock on a channel before attempting to work with it. Currently this is only implemented in the `existingChannelReducer`

lalexgap:
Closing this and will create a PR just for the storage component.

PR #601, opened 2019-07-09
Add path proposal

andrewgordstewart:


proposals/protocol-locator.md
```diff
@@ -0,0 +1,61 @@
...
+
+Think about some abstract "funding" protocol that both the server and the client need to coordinate on
+
+- Suppose the Foo protocol embeds two Bar protocols
+- They both need to be able to route actions to the "first" Bar protocol
+- Therefore, they agree that they will call that protocol "BarOne"
+- And they then decide to route actions to that protocol by making the protocolLocator start with `"BarOne"`
+
+The client is free to _store_ the protocols however it wants.
+However, I argue that it makes sense to store the "BarOne" protocol under the key `"BarOne"`, because that key describes what it is.
```
lalexgap: While I think the `protocolLocator` should look very similar to our `redux` state I don't think we want them directly coupled in any way. If a name of our redux state changes we shouldn't be forced to change the `protocolLocator`.

andrewgordstewart: When would the name in the redux state change, aside from a change in the protocol itself?

If the protocol changes, everything coordinating on that protocol needs to change.

lalexgap: I think it could happen in a refactor pretty easily.  Also I think a protocol locator like `fundingState/newLedgerFundingState` is less clear than `funding/newLedgerFunding`.

andrewgordstewart: I'm arguing that that refactor should _never_ happen, unless the protocol itself changed. I'm arguing that, by convention, in the client, in the redux store, that protocols are stored according to their locator.

andrewgordstewart: so, if `funding/newLedgerFunding` is more clear, then the funding state should have a property called `newLedgerFunding`, and not `newLedgerFundingState`

lalexgap: That would mean we would have to go through and rename all our current redux state, just so it looks nice on the `ProtocolLocator`. It also means whenever we're modifying state names we have to be mindful of how they look/handle with action routing. 

To me it seems that the `protocolLocator` and our redux store should be completely separate concerns.

andrewgordstewart: If you use this convention, assuming you store the path in an array, then you can do things like:
```
function someReducer(currentState, action) {
  const descriptor = path[0];
  const protocol = getProtocol(currentState, descriptor);
  return { ...currentState, [descriptor]: getReducer(descriptor)(currentState[descriptor], action)
}

function getProtocol(state, descriptor) {
  if ('descriptor' not in state) {
    // Handle the error somehow
  }

  return state.descriptor.protocol;
}

function getReducer(protocol: ProtocolName) {
  switch (protocol) {
    case 'INDIRECT_FUNDING':
      return indirectFundingReducer
    ...
    default:
      return unreachable(protocol)
  }
}
```

And you can do that in any protocol, with a commonly-defined `getReducer`. This means that handling of actions for embedded protocols is done in a completely conventional way, and you still have type safety.

tomclose:
Have you thought about whether each protocol should have a path _per participant_ vs overall? This seems like it would allow some flexibility in how a wallet wants to structure its internal state. I think mapping path to redux state might then be a sensible decision for our current browser wallet, but it would impose any restrictions on future wallets.

proposals/protocol-locator.md
andrewgordstewart: Maybe we are meaning two different things by a "protocol".

In the client wallet, we have a thing called a "protocol" which includes the definition of redux states, actions, reducers, etc.

But in the broader picture, there is a "funding protocol" that is a more abstract thing that both the client and the server (and who knows what else) needs to agree on as _the means of progressing a channel to a certain point_.

This abstract "funding protocol" needs to have some states like "WaitingForIndirectFunding", and some actions that specify state transitions.

> That would mean we would have to go through and rename all our current redux state, just so it looks nice on the `ProtocolLocator`. It also means whenever we're modifying state names we have to be mindful of how they look/handle with action routing.

Again, I'm arguing that we
A. first, describe what the protocols are. This should happen in code that is completely independent of the redux state, because this description is necessary for both the server and the client. The server cannot communicate with the client without knowing how the protocols work. Nor can an outside party build a wallet that communicates with our client.
B. second, structure the protocol states the natural way as dictated by the protocol descriptions.

Renaming variables in vscode is dead-easy, and shouldn't drive this decision. We're repeatedly blocked by not actually deciding how actions should be routed to embedded protocols, and that is much more costly than renaming variables. Each time we think about solving it, we come up with a peace-meal solution that
- works for the given situation
- does not take into consideration future complexity
- does not take into consideration the server
 
> To me it seems that the `protocolLocator` and our redux store should be completely separate concerns.

How the state looks should _not_ drive any decision on how the protocols look. Protocols need to be defined regardless of how things are stored in the client.

Once we've decided to use a tree structure data store (redux) to store a tree structure of objects (protocols), it seems totally senseless to me _not_ to use the same words to describe the branches of the two corresponding trees. If there is an actual benefit of not doing that, aside from not having to re-name variables, I'm all ears.

andrewgordstewart:
> Have you thought about whether each protocol should have a path _per participant_ vs overall? This seems like it would allow some flexibility in how a wallet wants to structure its internal state.

Can you give an example of what you mean by this? I think what you're suggesting would be akin to a url like
`${protocol}://${domain_name}/users/${user_id}`
rather than
`${protocol}://${domain_name}/users/?user_id=${user_id}`

Is that correct?

> I think mapping path to redux state might then be a sensible decision for our current browser wallet, but it would impose any restrictions on future wallets.

Choosing to map path to redux state is simply a decision that is sensible when storing a tree object in a tree data structure. If, in a future wallet, you are not storing your data in a tree, then you are free to store things however you want. Even in this case, the client wallet is free to store things however it wants, just like it does now -- it is totally irrelevant to the server what the redux state looks like.

But, as a developer, when
- the current state in protocol `protoA` has `path === 'some/location/description'`
- the current state embeds a protocol `protoB`
- the magmo-specification of `protoA` declares that the path for actions routed tot `protoB` is `'protoBState'`

then the convention of giving the redux state storing the state of `protoA`, a property called `protoBState`, that stores the protocol state of the `protoB`, _makes more sense than any other proposal I've seen_.

I am happy to have the flaws pointed out, but I don't know what your proposals are, and not solving this problem has blocked us several times already, and is blocking us yet again.

lalexgap:
Do we need to merge this in or is it obsoleted by #604?

PR #604, opened 2019-07-10
Protocol locator type guards

andrewgordstewart:
This PR changes type guards to be more strict, requiring that the next "entry" in the `protocolLocator` after the path it was given starts with the "descriptor". See the implementation of `isCommonAction`.

This is a draft PR since it would currently break the app: The reducers do not currently set the protocol locators. We should test the protocolLocator string as a part of `itSendsTheseCommitments`. To properly set the protocolLocator, one of two things is required:
A) A state must store its current "path". When a reducer of a state wants to send an action, it should append its path with the appropriate descriptor.
B) When a reducer calls a sub-reducer, and sees an outgoing message in the outbox, it pre-pends the `protocolLocator` with the appropriate description.

A seems like the simpler option.

In the meantime, I'd like some feedback about this `protocolLocator` design.


lalexgap:
This looks good to me. I'd vote for option A as it seems a lot cleaner than trying to modify outgoing messages.

packages/wallet/src/communication/actions.ts
```diff
@@ -133,3 +133,19 @@ export function isRelayableAction(action: WalletAction): action is RelayableActi
...
+export type CommonAction = CommitmentReceived | CommitmentsReceived;
+export function isCommonAction(
+  action: WalletAction,
+  path: string,
+  descriptor: string,
+): action is CommonAction {
+  return (
+    (action.type === 'WALLET.COMMON.COMMITMENTS_RECEIVED' ||
+      action.type === 'WALLET.COMMON.COMMITMENT_RECEIVED') &&
+    action.protocolLocator
```
lalexgap: We might want to just encapsulate this in a function since we'll be using it in a lot of type guards.

andrewgordstewart: This is the encapsulated function that's used in all the type guards 😛

packages/wallet/src/redux/protocols/advance-channel/reducer.ts
```diff
@@ -53,9 +53,10 @@ export const reducer: ProtocolReducer<states.AdvanceChannelState> = (
     console.error('Invalid action: expected WALLET.COMMON.COMMITMENTS_RECEIVED');
     return { protocolState, sharedData };
   }
-
-  if (action.type === 'WALLET.ADVANCE_CHANNEL.CLEARED_TO_SEND') {
-    return clearedToSendReducer(protocolState, sharedData);
+  {
```
lalexgap: Is there a reason the `if` statement is wrapped in `{`?

packages/wallet/src/redux/protocols/virtual-funding/reducer.ts
```diff
@@ -34,7 +40,7 @@ export function initialize(sharedData: SharedData, args: states.InitializationAr
     commitmentType: CommitmentType.PreFundSetup,
     clearedToSend: true,
     processId,
-    protocolLocator: states.GUARANTOR_CHANNEL_DESCRIPTOR,
+    protocolLocator: ADVANCE_CHANNEL_PROTOCOL_LOCATOR,
```
lalexgap: Should this be `makeLocator(EmbeddedProtocol.VirtualFunding,EmbeddedProtocol.AdvanceChannel)`?

packages/wallet/src/redux/protocols/funding/player-b/reducer.ts
```diff
@@ -56,10 +55,7 @@ export function fundingReducer(
   sharedData: SharedData,
   action: actions.FundingAction | EmbeddedAction,
 ): ProtocolStateWithSharedData<states.FundingState> {
-  if (
-    isAdvanceChannelAction(action) &&
-    action.protocolLocator === ADVANCE_CHANNEL_PROTOCOL_LOCATOR
-  ) {
+  if (isAdvanceChannelAction(action)) {
```
lalexgap: I think we need to check the `protocolLocator` in addition to checking `isAdvanceChannelAction` since an `AdvanceChannelAction`  could be meant this protocol or a sub protocol. Without checking the `protocolLocator`we'll prevent any `AdvanceChannelAction`s from being sent to a sub protocol.

packages/wallet/src/redux/protocols/funding/player-a/reducer.ts
```diff
@@ -58,11 +57,7 @@ export function fundingReducer(
   sharedData: SharedData,
   action: actions.FundingAction | EmbeddedAction,
 ): ProtocolStateWithSharedData<states.FundingState> {
-  if (
-    isAdvanceChannelAction(action) &&
-    // TODO: Remove this check once protocol-locator updates have been made
-    action.protocolLocator === ADVANCE_CHANNEL_PROTOCOL_LOCATOR
-  ) {
+  if (isAdvanceChannelAction(action)) {
```
lalexgap: Same issue as mentioned above.

packages/wallet/src/redux/protocols/new-ledger-funding/reducer.ts
```diff
@@ -118,11 +121,7 @@ function handleWaitForPostFundSetup(
       },
       sharedData,
     };
-  } else if (
-    isAdvanceChannelAction(action) &&
-    // TODO: Remove this check once the protocol-locator has been properly implemented.
-    action.protocolLocator === NEW_LEDGER_FUNDING_PROTOCOL_LOCATOR
-  ) {
+  } else if (isAdvanceChannelAction(action)) {
```
lalexgap: We might want to check the `ProtocolLocator` here as well.

packages/wallet/src/redux/protocols/new-ledger-funding/states.ts
```diff
@@ -100,6 +99,14 @@ export function isNewLedgerFundingState(state: ProtocolState): state is NewLedge
...
+  return state.type === 'NewLedgerFunding.Success';
+}
+
+export function isFailure(state: NewLedgerFundingState): state is Failure {
+  return state.type === 'NewLedgerFunding.Failure';
+}
+
 export function isTerminal(state: NewLedgerFundingState): state is Failure | Success {
-  return state.type === 'NewLedgerFunding.Failure' || state.type === 'NewLedgerFunding.Success';
+  return !![isSuccess, isFailure].find(g => g(state));
```
lalexgap: Minor but I think `return isSuccess(state) || isFailure(state)` is a lot clearer :P

PR #605, opened 2019-07-10
Exchange Post Fund Setups in Funding/NewLedgerFunding

### LINKED ISSUE #593: Exchange PostFundSetups in the Funding/NewLedgerFunding protocols

lalexgap:
To support Virtual funding post fund setup exchange has been moved to the `funding` protocol (for application channels) and the `newLedgerFunding` protocol (for ledger channels).

The `NewLedgerFunding` protocol has also been refactored into one reducer (from a reducer for player A and player B) and now makes use of the `advanceChannel` and `consensusUpdate` protocol.

Fixes #593


packages/wallet/src/redux/protocols/funding/player-a/reducer.ts
```diff
@@ -43,7 +53,12 @@ export function fundingReducer(
   sharedData: SharedData,
   action: actions.FundingAction | EmbeddedAction,
 ): ProtocolStateWithSharedData<states.FundingState> {
-  if (isIndirectFundingAction(action)) {
+  if (
+    isAdvanceChannelAction(action) &&
+    action.protocolLocator === ADVANCE_CHANNEL_PROTOCOL_LOCATOR
```
andrewgordstewart:
```suggestion
    // TODO: Remove this check once protocol-locator updates have been made
    action.protocolLocator === ADVANCE_CHANNEL_PROTOCOL_LOCATOR
```

packages/wallet/src/redux/protocols/new-ledger-funding/__tests__/scenarios.ts
```diff
@@ -0,0 +1,130 @@
...
+    channelFromCommitments([ledger4, ledger5], asAddress, asPrivateKey),
+  ]),
+};
+
+const waitForDirectFundingFailure = states.waitForDirectFunding({
+  ...props,
+  directFundingState: DFPreFailure.state,
+  postFundSetupState: ACPreSuccess.state,
+});
+// Since we rely extensively on sub-protocols
```
andrewgordstewart: Outside the scope of this PR, but it might be worth configuring the `preSuccess` exports in these dependencies to be a function of the player index.

packages/wallet/src/redux/protocols/new-ledger-funding/readme.md
```diff
@@ -7,18 +7,10 @@ The new ledger funding protocol coordinates the process of funding an applicatio
...
 ## The Protocol
 
-The indirect funding protocol involves two parties: player A and player B.
-Player A is identified by being the first participant in the `participants` array in
-the application channel X that is to be funded.
-
-We therefore split the overall indirect-funding protocol into two sub-protocols: the
-[player-a-indirect-funding protocol](./player-a/readme.md) and the [player-b-indirect-funding protocol](./player-b/readme.md).
-
 The two protocols interact through the following messages:
```
andrewgordstewart:
```suggestion
The protocol interacts between client wallets through the following messages:
```

packages/wallet/src/redux/protocols/new-ledger-funding/reducer.ts
```diff
@@ -1,42 +1,385 @@
...
+type IDFAction = NewLedgerFundingAction;
+export const NEW_LEDGER_FUNDING_PROTOCOL_LOCATOR = 'NewLedgerFunding';
 export function initialize(
   processId: string,
-  channel: ChannelState,
+  channelId: string,
+  sharedData: SharedData,
+): ReturnVal {
+  const privateKey = getPrivatekey(sharedData, channelId);
+  const ourIndex = isFirstPlayer(channelId, sharedData)
```
andrewgordstewart: This could be replaced with a `getPlayerIndex(channelId, sharedData)` helper.

packages/wallet/src/redux/protocols/new-ledger-funding/reducer.ts
```diff
@@ -1,42 +1,385 @@
...
+}
+
+function handleWaitForDirectFunding(
+  protocolState: states.WaitForDirectFunding,
+  sharedData: SharedData,
+  action: IDFAction | DirectFundingAction | ConsensusUpdateAction,
+): ReturnVal {
+  if (
+    isAdvanceChannelAction(action) &&
+    action.protocolLocator === NEW_LEDGER_FUNDING_PROTOCOL_LOCATOR
```
andrewgordstewart: This check seems wrong. The protocolLocator should be used to determine where the action goes, so in this case, the protocolLocator should be `ADVANCE_CHANNEL_PROTOCOL_LOCATOR` if it should be routed to the advanceChannel protocol.

packages/wallet/src/redux/protocols/new-ledger-funding/reducer.ts
```diff
@@ -1,42 +1,385 @@
...
+  allocation: string[],
+  destination: string[],
+): { allocation: string[]; destination: string[]; appAttributes: string } {
+  return {
+    allocation,
+    destination,
+    appAttributes: bytesFromAppAttributes({
+      proposedAllocation: allocation,
+      proposedDestination: destination,
+      furtherVotesRequired: 0,
```
andrewgordstewart: I think this should be the number of participants, rather than `0`.

lalexgap: I think you're right. This should probably be changed in `virtualFunding` as well (where I copied it from)

packages/wallet/src/redux/protocols/new-ledger-funding/reducer.ts
lalexgap: Right now we can't just use `ADVANCE_CHANNEL_PROTOCOL_LOCATOR` as the `funding` protocol also uses `advanceChannel` so the `funding` protocol isn't able to determine if an `advanceChannel` action is for itself or `newLedgerFunding`. I think this will be resolved when we get `protocolLocator` working correctly.

packages/wallet/src/redux/protocols/new-ledger-funding/reducer.ts
```diff
@@ -1,42 +1,385 @@
...
+    return {
+      protocolState: {
+        ...protocolState,
+        consensusUpdateState: consensusUpdateResult.protocolState,
+      },
+      sharedData,
+    };
+  } else if (
+    isAdvanceChannelAction(action) &&
+    action.protocolLocator === NEW_LEDGER_FUNDING_PROTOCOL_LOCATOR
```
andrewgordstewart:
```suggestion
    // TODO: Remove this check once the protocol-locator has been properly implemented.
    action.protocolLocator === NEW_LEDGER_FUNDING_PROTOCOL_LOCATOR
```

PR #607, opened 2019-07-10
Get Funding working

### LINKED ISSUE #606: Funding currently fails with RPS

lalexgap:
A small change broke funding by starting ledger channels with a propose commitment instead of a consensus commitment.

Fixes #606


lalexgap:
Still running into this.

lalexgap:
There is still a race condition but I've created #609 to handle that.

lalexgap:
> Is the issue that `furtherVotesRequired` set in postfund commitments gets inherited by the first application commitment?

Yup, the issue is creating the first proposal fails if `furtherVotesRequired` is not 0, since if `furtherVotesRequired` is not 0 it is not a consensus commitment.

PR #623, opened 2019-07-19
Correctly set funding state in New Ledger Channel protocol

### LINKED ISSUE #622: Wallet always creates a new Ledger Channel for funding
After the latest changes the wallet seems to always create a new ledger channel to fund a game instead of using an existing one.

lalexgap:
Funding state was not being set correctly so existing ledger channels were not being found when trying to fund the game with an existing ledger channel.

Fixes #622

### DIFF
--- a/packages/wallet/src/redux/protocols/indirect-funding/__tests__/scenarios.ts
+++ b/packages/wallet/src/redux/protocols/indirect-funding/__tests__/scenarios.ts
@@ -15,6 +15,7 @@ import { preSuccess as existingLedgerPreSuccess } from '../../existing-ledger-fu
 import {
   preSuccessState as newLedgerPreSuccess,
   successTrigger as NewLedgerChannelSuccessTrigger,
+  successState as newLedgerSuccess,
 } from '../../new-ledger-channel/__tests__';
 import { INDIRECT_FUNDING_PROTOCOL_LOCATOR } from '../reducer';
 
@@ -36,7 +37,7 @@ const ledger4 = ledgerCommitment({ turnNum: 4, balances: oneThree });
 const ledger5 = ledgerCommitment({ turnNum: 5, balances: oneThree });
 const app0 = appCommitment({ turnNum: 0, balances: oneThree });
 const app1 = appCommitment({ turnNum: 1, balances: oneThree });
-const existingLedgerFundingSharedData = setChannels(existingLedgerPreSuccess.sharedData, [
+const existingLedgerFundingSharedData = setChannels(newLedgerSuccess.store, [
   channelFromCommitments([ledger4, ledger5], asAddress, asPrivateKey),
   channelFromCommitments([app0, app1], asAddress, asPrivateKey),
 ]);
--- a/packages/wallet/src/redux/protocols/new-ledger-channel/reducer.ts
+++ b/packages/wallet/src/redux/protocols/new-ledger-channel/reducer.ts
@@ -309,7 +309,6 @@ function channelSpecificArgs(
 }
 
 function updateFundingState(sharedData: SharedData, appChannelId: string, ledgerId: string) {
-  // update fundingState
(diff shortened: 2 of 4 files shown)


packages/wallet/src/redux/protocols/new-ledger-channel/reducer.ts
```diff
@@ -301,3 +307,17 @@ function channelSpecificArgs(
     }),
   };
 }
+
+function updateFundingState(sharedData: SharedData, appChannelId: string, ledgerId: string) {
+  // update fundingState
```
andrewgordstewart:
```suggestion

```

packages/wallet/src/redux/protocols/new-ledger-channel/reducer.ts
```diff
@@ -301,3 +307,17 @@ function channelSpecificArgs(
...
+  const channelFundingState: ChannelFundingState = {
+    directlyFunded: false,
+    fundingChannel: ledgerId,
+  };
+  const ledgerFundingState: ChannelFundingState = {
+    directlyFunded: true,
+  };
+  sharedData = setFundingState(sharedData, appChannelId, channelFundingState);
```
andrewgordstewart: If we are overwriting the existing reference to `sharedData` (which could have side effects in some other code) is it  worth just overwriting that reference in `setFundingState`?

If that's the case, `updateFundingState` can return `void`, and no caller needs to remember to return the new, updated shared data.

packages/wallet/src/redux/protocols/indirect-funding/__tests__/scenarios.ts
```diff
@@ -36,7 +37,7 @@ const ledger4 = ledgerCommitment({ turnNum: 4, balances: oneThree });
 const ledger5 = ledgerCommitment({ turnNum: 5, balances: oneThree });
 const app0 = appCommitment({ turnNum: 0, balances: oneThree });
 const app1 = appCommitment({ turnNum: 1, balances: oneThree });
-const existingLedgerFundingSharedData = setChannels(existingLedgerPreSuccess.sharedData, [
+const existingLedgerFundingSharedData = setChannels(newLedgerSuccess.store, [
```
andrewgordstewart: This line seems strange. Should the existingLedgerFunding preSucces instead use the newLedgerSuccess store to generate its store?

packages/wallet/src/redux/protocols/new-ledger-channel/reducer.ts
```diff
@@ -301,3 +307,17 @@ function channelSpecificArgs(
     }),
   };
 }
+
+function updateFundingState(sharedData: SharedData, appChannelId: string, ledgerId: string) {
+  // update fundingState
+  const channelFundingState: ChannelFundingState = {
```
andrewgordstewart: :+1:  for types!

packages/wallet/src/redux/protocols/indirect-funding/__tests__/scenarios.ts
lalexgap: Hmm, I think I'd lean towards leaving it this way since relying on `existingLedgerPreSuccess` seems less accurate.

We're trying to set up the `sharedData` before the existing ledger protocol runs but by using `existingLedgerPreSuccess` we're using `sharedData` of the existing ledger channel protocol that is partially completed.

This seems to be closer to what actually happens, since the existing ledger channel protocol runs after the new ledger channel protocol has succeeded in the past.

packages/wallet/src/redux/protocols/new-ledger-channel/reducer.ts
lalexgap: Overwriting the `sharedData` parameter doesn't modify it outside the scope of the function.

PR #626, opened 2019-07-19
Handle Commitment out of order in Concluding

### LINKED ISSUE #624: Concluding occasionally stalls waiting for Consensus Update
Concluding can stall on consensus update. It only seems to happen under certain conditions.

lalexgap:
Handles a consensus update commitment coming in early.

Fixes #624

### DIFF
--- a/packages/wallet/src/redux/protocols/concluding/instigator/reducer.ts
+++ b/packages/wallet/src/redux/protocols/concluding/instigator/reducer.ts
@@ -138,10 +138,10 @@ function handleLedgerUpdateAction(
     };
   } else {
     return {
-      protocolState: states.instigatorWaitForLedgerUpdate({
+      protocolState: {
         ...protocolState,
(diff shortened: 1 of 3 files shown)


packages/wallet/src/redux/protocols/concluding/instigator/reducer.ts
```diff
@@ -138,10 +138,10 @@ function handleLedgerUpdateAction(
     };
   } else {
     return {
-      protocolState: states.instigatorWaitForLedgerUpdate({
```
geoknee: This change seems to make sense, but when I tried reversing it, the new test still passed. Am I missing something?

lalexgap: Oops, I was using the wrong action in the scenario. Good catch! Fixed in f92cb9a8bb2bf42b6e0d864ae5b7b58d987c4178

PR #628, opened 2019-07-20
Fix Ledger Top Up

### LINKED ISSUE #625: Ledger Top Up protocol doesn't work
The Ledger top up protocol doesn't seem to work after the latest changes.

lalexgap:
Two main changes in this PR:
- Refactor Existing Ledger Funding to use Consensus Update
- Start Consensus Update protocol early in Ledger Top Up to handle commitments that might come in before we get a `FUNDING_RECEIVED` event.

Fixes #625

### DIFF
--- a/packages/wallet/src/redux/protocols/existing-ledger-funding/__tests__/reducer.test.ts
+++ b/packages/wallet/src/redux/protocols/existing-ledger-funding/__tests__/reducer.test.ts
@@ -146,20 +146,3 @@ function itTransitionsTo(state: ReturnVal, type: states.ExistingLedgerFundingSta
     expect(state.protocolState.type).toEqual(type);
   });
 }
-
-// function itSendsMessage(state: ReturnVal, message: SignedCommitment) {
-//   it('sends a message', () => {
-//     const lastMessage = getLastMessage(state.sharedData);
-//     if (lastMessage && 'messagePayload' in lastMessage) {
-//       const dataPayload = lastMessage.messagePayload;
-//       // This is yuk. The data in a message is currently of 'any' type..
-//       if (!('signedCommitment' in dataPayload)) {
-//         fail('No signedCommitment in the last message.');
-//       }
-//       const { commitment, signature } = dataPayload.signedCommitment;
-//       expect({ commitment, signature }).toEqual(message);
-//     } else {
-//       fail('No messages in the outbox.');
-//     }
-//   });
-// }
--- a/packages/wallet/src/redux/protocols/existing-ledger-funding/__tests__/reducer.test.ts
+++ b/packages/wallet/src/redux/protocols/existing-ledger-funding/__tests__/reducer.test.ts
@@ -2,8 +2,8 @@ import * as scenarios from './scenarios';
 import { initialize, existingLedgerFundingReducer } from '../reducer';
 import * as states from '../states';
 import { ProtocolStateWithSharedData } from '../..';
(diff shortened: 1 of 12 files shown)


packages/wallet/src/redux/protocols/existing-ledger-funding/__tests__/reducer.test.ts
```diff
@@ -67,10 +66,9 @@ describe('player B happy path', () => {
   });
 
   describeScenarioStep(scenario.waitForLedgerUpdate, () => {
-    const { state, action, sharedData, reply } = scenario.waitForLedgerUpdate;
+    const { state, action, sharedData } = scenario.waitForLedgerUpdate;
     const updatedState = existingLedgerFundingReducer(state, sharedData, action);
     itTransitionsTo(updatedState, 'ExistingLedgerFunding.Success');
-    itSendsMessage(updatedState, reply);
```
lalexgap: Now that we're using the consensus update protocol we can't really expect a certain reply in our tests. This is due to the fact that Consensus Update will reply with whatever its success commitment is which can differ from what we expect here.

We could just check that some message is sent out or start exporting a `successReply` from Consensus Update?

packages/wallet/src/redux/protocols/indirect-funding/reducer.ts
```diff
@@ -4,11 +4,12 @@ import * as selectors from '../../selectors';
 import * as helpers from '../reducer-helpers';
 import { getLastCommitment, ChannelState } from '../../channel-store/channel-state';
 import { CommitmentType } from 'fmg-core';
+import { isExistingLedgerFundingAction } from '../existing-ledger-funding';
+// TODO: Why does importing the reducer from the index result in test failures in grand parent protocols?
```
lalexgap: Not sure why but importing from the index file caused this error ` TypeError: existing_ledger_funding_1.existingLedgerFundingReducer is not a function`  in a couple tests.

packages/wallet/src/redux/protocols/existing-ledger-funding/__tests__/reducer.test.ts
andrewgordstewart: I think exporting a `successReply` is a better idea than not testing it.

packages/wallet/src/redux/protocols/indirect-funding/reducer.ts
andrewgordstewart: I actually got `TypeError: existing_ledger_funding_2.initializeExistingLedgerFunding is not a function` (note the `_2`, not `_1`).

Strange.

packages/wallet/src/redux/protocols/ledger-top-up/reducer.ts
```diff
@@ -195,45 +199,47 @@ const waitForDirectFundingForAReducer: ProtocolReducer<states.LedgerTopUpState>
...
-    const newProtocolState = states.restoreOrderAndAddBTopUpUpdate({
-      ...protocolState,
-      consensusUpdateState,
-    });
-    return {
-      protocolState: newProtocolState,
-      sharedData: newSharedData,
-    };
+    const { protocolLocator, processId } = protocolState;
+    if (directFundingState.type === 'DirectFunding.FundingFailure') {
```
andrewgordstewart: Maybe `switch` on `directFundingState.type`? I think it's more readable and also more robust, if we use `unreachable`.

Issue #629, opened 2019-07-20
Error handling when loading/saving to storage


lalexgap:
We should add error handling when saving/loading from storage, to prevent storage errors from breaking the whole app.

PR #630, opened 2019-07-20
Implement redux storage

### LINKED ISSUE #22: Investigate the use of IndexedDB to store state

lalexgap:
This introduces `redux-storage` and stores channel info (and some related info) in storage.

We load from indexedDB when the page is loaded and we save to indexedDB after every action(this can be configured to a specific list of actions). Currently there's no error handling for storage related failures(#629).

Instead of storing `adjudicatorState` there is saga that updates `adjudicatorState` based on the latest on-chain info.

A quick way of seeing storage working is to
- Open a channel, play a game, and select `Keep Open` when concluding
- Refresh one more browsers
- Open a new channel. It will use the existing ledger channel that was saved in storage.

Fixes #22.


packages/wallet/src/redux/reducer.ts
```diff
@@ -201,7 +201,13 @@ const waitForLoginReducer = (
 ): states.WalletState => {
   switch (action.type) {
     case 'WALLET.LOGGED_IN':
-      const { privateKey, address } = ethers.Wallet.createRandom();
+      let address = state.address;
+      let privateKey = state.privateKey;
+      if (!address || !privateKey) {
+        const { privateKey: newPrivateKey, address: newAddress } = ethers.Wallet.createRandom();
```
andrewgordstewart:
```suggestion
        ({ privateKey, address } = ethers.Wallet.createRandom());
```

packages/wallet/src/redux/reducer.ts
```diff
@@ -201,7 +201,13 @@ const waitForLoginReducer = (
...
+        address = newAddress;
```
andrewgordstewart:
```suggestion
```

packages/wallet/src/redux/reducer.ts
```diff
@@ -201,7 +201,13 @@ const waitForLoginReducer = (
...
+        privateKey = newPrivateKey;
```
andrewgordstewart:
```suggestion
```

packages/wallet/src/redux/reducer.ts
```diff
@@ -201,7 +201,13 @@ const waitForLoginReducer = (
 ): states.WalletState => {
   switch (action.type) {
     case 'WALLET.LOGGED_IN':
-      const { privateKey, address } = ethers.Wallet.createRandom();
+      let address = state.address;
```
andrewgordstewart:
```suggestion
      let { address, privateKey } = state;
```

packages/wallet/src/redux/reducer.ts
```diff
@@ -201,7 +201,13 @@ const waitForLoginReducer = (
...
+      let privateKey = state.privateKey;
```
andrewgordstewart:
```suggestion
```

packages/wallet/src/redux/adjudicator-state/reducer.ts
```diff
@@ -59,6 +61,14 @@ const fundingReceivedEventReducer = (
   const { channelId } = action;
   return setBalance(state, channelId, action.totalForDestination);
 };
+const channelUpdateReducer = (state: AdjudicatorState, action: actions.ChannelUpdate) => {
```
andrewgordstewart: Is this related to storing the redux-store in storage?

packages/wallet/src/redux/state.ts
```diff
@@ -56,6 +56,8 @@ export interface ChannelSubscriptions {
 }
 export interface WaitForLogin extends SharedData {
   type: typeof WAIT_FOR_LOGIN;
+  address?: string;
```
andrewgordstewart: I'm wondering if it's better not to store the address, which is a pure function of the private key.

packages/wallet/src/redux/adjudicator-state/reducer.ts
lalexgap: Yes, it's used to update the `adjudicator-state` on load for stored channels. We currently don't store the `adjudicator-state` so we need to initialize it on load from storage.

packages/wallet/src/redux/state.ts
lalexgap: Going to leave this as is for now to get the PR in, but perhaps something we can consider in the future?

packages/wallet/src/redux/store.ts
```diff
@@ -1,16 +1,33 @@
...
 
 import { walletReducer } from './reducer';
 import { sagaManager } from './sagas/saga-manager';
+import filter from 'redux-storage-decorator-filter';
+import createEngine from 'redux-storage-engine-indexed-db';
+
+// We currently whitelist the values that we store/load.
+const storageEngine = filter(createEngine('magmo-wallet'), [
+  'whitelisted-key',
+  ['address'],
```
andrewgordstewart: A comment that didn't seem to make it into my review. This part is tightly coupled with the shape of the shape of the redux store. Perhaps the whitelisted keys should be defined next to that interface and imported here?

PR #633, opened 2019-07-22
Offer choice of virtual or indirect funding when funding application channel

andrewgordstewart:
The UI doesn't allow a choice, and only lets the user approve/reject virtual funding. The protocol, however, does support choice.


packages/wallet/src/redux/protocols/funding/player-a/reducer.ts
```diff
@@ -165,47 +203,102 @@ function strategyChosen(
...
+        targetChannelId,
+        ourIndex,
+        // TODO: This should be an env variable
+        hubAddress: '0x100063c326b27f78b2cBb7cd036B8ddE4d4FCa7C',
+        startingAllocation,
+        startingDestination,
+        protocolLocator: makeLocator(EmbeddedProtocol.VirtualFunding),
+      });
+
+      const advanceChannelResult = initializeAdvanceChannel(
```
lalexgap: Could this be pulled out of the switch so it's not duplicated?

packages/wallet/src/redux/protocols/funding/player-a/__tests__/scenarios.ts
```diff
@@ -22,7 +23,7 @@ import { EmbeddedProtocol } from '../../../../../communication';
 //
 // 1. Happy path: WaitForStrategyChoice
```
lalexgap: Could we update the comment with the new scenario? (Or just remove it, I don't think it's really necessary)

andrewgordstewart: I think these comments are unnecessary, and are just a pain to keep up to date!

PR #642, opened 2019-07-24
Valid transition

andrewgordstewart:
- updates fmg-nitro-adjudicator to v0.1.5, which implements `validTransition` (with a funky extra argument)
- rethrows unhandled errors when `validTransition` throws


packages/server/package.json
```diff
@@ -17,7 +17,7 @@
     "chain:test": "npx ganache-then-jest --runInBand -c ./config/jest/jest.chain.config.js",
     "adjudicator-watcher:start": "node lib/wallet/adjudicator-watcher/index.js",
     "adjudicator-watcher:test": "npx ganache-then-jest --runInBand -c ./config/jest/jest.chain.config.js",
-    "firebase-relay:start:dev": "ts-node src/message/firebase-relay",
+    "firebase-relay:start:dev": "nodemon --watch 'src/**/*' -e ts,tsx --exec ts-node src/message/firebase-relay",
```
lalexgap: Should this be part of this PR?

andrewgordstewart: It's kind of unrelated, but it came in handy when debugging so I'd vote to keep it.

PR #643, opened 2019-07-24
Separate out Messaging from Response logic

lalexgap:
I've attempted to separate out the logic of message handling (from either firebase or the REST endpoint) from the wallet/rps logic.

`firebaseRelay` no longer relies on posting to the REST endpoint and having the REST endpoint call `sendToFirebase`


lalexgap:
Looks like there's an issue with the firebase relay as it doesn't have a connection to the database.

packages/server/src/app/handlers/handle-wallet-message.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (await isNewProcessAction(message)) {
+    return handleNewProcessAction(message);
+  } else if (await isProtocolAction(message)) {
+    return handleOngoingProcessAction(message);
+  } else {
+    return undefined;
+  }
+}
+
+function isNewProcessAction(action: RelayableAction): boolean {
```
andrewgordstewart: if this is a type guard, then `handleNewProcessAction` can accept only the `CONCLUDE_INSTIGATED` type, which simplifies the implementation of `handleNewProcessAction`

packages/server/src/app/handlers/handle-wallet-message.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (action.type === 'WALLET.NEW_PROCESS.CONCLUDE_INSTIGATED') {
+    return true;
+  } else {
+    return false;
+  }
+}
+
+async function isProtocolAction(action: RelayableAction): Promise<boolean> {
```
andrewgordstewart: Again, perhaps this should be a type guard. (We can move the async code to the block at L13.

packages/server/src/message/firebase-relay.ts
```diff
@@ -27,31 +34,29 @@ function getMessagesRef() {
...
   const hubRef = getMessagesRef().child(HUB_ADDRESS.toLowerCase());
 
-  hubRef.on('child_added', snapshot => {
+  hubRef.on('child_added', async snapshot => {
     const key = snapshot.key;
     const value = snapshot.val();
     const queue = value.queue;
     if (queue === 'GAME_ENGINE') {
-      postToHub(value);
+      const outgoingMessage = await handleAppMessage(value);
```
andrewgordstewart: 💡 throw an error here to be explicit about the fact that the current server does not support hub-play.

packages/server/src/message/firebase-relay.ts
```diff
@@ -1,8 +1,15 @@
...
+import { Model } from 'objection';
 import '../../config/env';
+import { handleAppMessage } from '../app/handlers/handle-app-message';
+import { handleWalletMessage } from '../app/handlers/handle-wallet-message';
 import { HUB_ADDRESS } from '../constants';
+import knex from '../wallet/db/connection';
+
+// TODO: Currently the firebaseRelay is all that is needed to respond to the app
+// If the REST endpoint is not being used it should probably be removed
+Model.knex(knex);
```
andrewgordstewart: The firebase-relay shouldn't be responsible for setting up a db connection. As per our discussion, there should be a `hub` that has a connection, and the firebase relay can call `hub.handleWalletMessage`.

PR #644, opened 2019-07-24
Update player B container

lalexgap:
The player B container was still dispatching an action with `IndirectFundingStrategy` instead of `Virtual`


lalexgap:
Wrong branch

PR #647, opened 2019-07-25
Fix firebase sending

lalexgap:
Fixed the firebase relay to properly send the messages

### DIFF
--- a/packages/server/src/hub/server.ts
+++ b/packages/server/src/hub/server.ts
@@ -21,13 +21,15 @@ const firebaseRelay = fork(`${__dirname}/../message/firebase-relay`, [], {
   execArgv: forkExecArgv,
 });
 firebaseRelay.on('message', (message: RelayableAction) => {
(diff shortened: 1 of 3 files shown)


packages/server/src/hub/server.ts
```diff
@@ -21,13 +21,15 @@ const firebaseRelay = fork(`${__dirname}/../message/firebase-relay`, [], {
   execArgv: forkExecArgv,
 });
 firebaseRelay.on('message', (message: RelayableAction) => {
-  console.log(`Parent process received message from firebase": ${JSON.stringify(message)}`);
+  console.log(
+    `Parent process received message from firebase": ${JSON.stringify(message, null, 1)}`,
```
andrewgordstewart: What's the significance of `1`?

lalexgap: `1` is the number of spaces to use when printing out the JSON object. So instead of a blob it gets printed out nicely formatted.

PR #648, opened 2019-07-25
Get Virtual Funding Partially Working

lalexgap:
This is work towards getting virtual funding working. Currently virtual funding gets to the point of attempting to fund the ledger channel that funds the guarantor channel.

Things are at a mostly stable point so might as well merge this in so the PR doesn't get too unwieldy.


packages/server/src/wallet/db/queries/channels.ts
```diff
@@ -64,21 +64,30 @@ async function updateAllocatorChannel(
...
+    nonce,
+    guaranteedChannel,
+  };
 
-  // For now, we just _assume_ that the channel is fully funded
-  const holdings = allocations(hubCommitment)
-    .map(x => x.amount)
-    .reduce((a, b) =>
+  // TODO: We are currently using the allocations to set the funding amount
+  // This assumes that the channel is funded and DOES NOT work for guarantor channels
```
andrewgordstewart: I'm not sure I understand this comment.

packages/server/src/wallet/db/queries/channels.ts
```diff
@@ -64,21 +64,30 @@ async function updateAllocatorChannel(
...
+  const hubAllocationAmounts = allocations(hubCommitment).map(x => x.amount);
+  let holdings = bigNumberify(0).toHexString();
+  if (hubAllocationAmounts.length > 0) {
+    holdings = hubAllocationAmounts.reduce((a, b) =>
```
andrewgordstewart: If you pass the optional initial value to `reduce`, you can make `holdings` a constant, and remove the check on `hubAllocationAmounts.length > 0`

`hubAllocationAmounts.reduce( [callback in L83-86], bigNumberify(0).toHexString())`

packages/server/src/hub/handlers/handle-wallet-message.ts
```diff
@@ -13,23 +13,33 @@ import { handleOngoingProcessAction } from './handle-ongoing-process-action';
...
 
-function isNewProcessAction(action: RelayableAction): action is ConcludeInstigated {
-  return action.type === 'WALLET.NEW_PROCESS.CONCLUDE_INSTIGATED';
+async function shouldHandleAsNewProcessAction(
+  action: ConcludeInstigated | CommitmentsReceived,
+): Promise<boolean> {
+  if (action.type !== 'WALLET.COMMON.COMMITMENTS_RECEIVED') {
+    return true;
+  }
+  return !(await getProcess(action.processId));
```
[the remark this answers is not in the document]
lalexgap: The only downside of that approach is that if for some reason the client decides to broadcast commitments instead of send them serially then this check won't work.

packages/server/src/wallet/db/queries/channels.ts
lalexgap: Right now we assume the amount of funding a channel has is the sum of its `allocation`s. This assumes the channel is fully funded. 

For guarantor channels since there's no 'allocation's `holdings` will always be 0 no matter the amount of funds deposited into that channel.

PR #650, opened 2019-07-30
Get Virtual Funding Working

lalexgap:
Gets virtual funding to the point of being able to actually fund a game of RPS.

### DIFF
--- a/packages/server/src/hub/handlers/handle-ongoing-process-action.ts
+++ b/packages/server/src/hub/handlers/handle-ongoing-process-action.ts
@@ -138,19 +138,16 @@ async function handleCommitmentsReceived(action: CommitmentsReceived) {
       ledgerCommitmentRound,
       currentCommitment && asConsensusCommitment(currentCommitment),
     );
-    const response = [];
-    const otherParticipants = participants.filter(p => participants.indexOf(p) !== ourIndex);
-    for (const participant of otherParticipants) {
-      response.push(
+    return participants
+      .filter((_, idx) => idx !== ourIndex)
+      .map(p =>
         communication.sendCommitmentsReceived(
-          participant,
+          p,
           processId,
           [...incomingCommitments, { commitment, signature: (signature as unknown) as string }],
           action.protocolLocator,
         ),
       );
-    }
-    return response;
   }
 }
 
--- a/packages/server/src/hub/handlers/handle-ongoing-process-action.ts
+++ b/packages/server/src/hub/handlers/handle-ongoing-process-action.ts
@@ -21,7 +21,7 @@ import { updateRPSChannel } from '../services/rpsChannelManager';
 
 export async function handleOngoingProcessAction(
   action: StrategyProposed | CommitmentReceived | CommitmentsReceived,
-): Promise<MessageRelayRequested | undefined> {
+): Promise<MessageRelayRequested[]> {
   switch (action.type) {
     case 'WALLET.COMMON.COMMITMENT_RECEIVED':
       return handleCommitmentReceived(action);
@@ -42,7 +42,7 @@ async function handleStrategyProposed(action: StrategyProposed) {
   }
 
   const { theirAddress } = process;
-  return communication.sendStrategyApproved(theirAddress, processId, strategy);
+  return [communication.sendStrategyApproved(theirAddress, processId, strategy)];
 }
 
 async function handleCommitmentReceived(action: CommitmentReceived) {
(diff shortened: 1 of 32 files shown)


packages/server/src/hub/handlers/handle-ongoing-process-action.ts
```diff
@@ -132,12 +138,19 @@ async function handleCommitmentsReceived(action: CommitmentsReceived) {
...
     );
-    return communication.sendCommitmentsReceived(
-      nextParticipant,
-      processId,
-      [...incomingCommitments, { commitment, signature: (signature as unknown) as string }],
-      action.protocolLocator,
-    );
+    const response = [];
+    const otherParticipants = participants.filter(p => participants.indexOf(p) !== ourIndex);
+    for (const participant of otherParticipants) {
```
lalexgap: This is a work around to the #651. 

Having the hub broadcast messages to all participants is a simple fix but it ties us to a model where the hub can message every participant (IE: not a REST endpoint)

packages/server/src/wallet/services/blockchain.ts
```diff
@@ -1,16 +1,20 @@
...
 
+import * as AsyncLock from 'async-lock';
 import { nitroAdjudicator } from '../utilities/blockchain';
-
+const lock = new AsyncLock();
 export class Blockchain {
   static nitro: any;
   static async fund(channelID: Address, expectedHeld: Uint256, value: Uint256): Promise<Uint256> {
-    await Blockchain.attachNitro();
+    // We lock to prevent issues with the transaction nonce getting out of sync in ethers.js
```
andrewgordstewart: Can you explain how this kind of out-of-sync can arise? This kind of explanation is valuable to have in the commit message.

packages/server/src/hub/handlers/handle-ongoing-process-action.ts
andrewgordstewart: This can be
```
const response = participants
  .filter((_, idx) => idx !== ourIndex)
  .map(p => communication.sendCommitmentsReceived(...))
```

packages/wallet/src/contract-tests/adjudicator-events.test.ts
```diff
@@ -20,7 +20,12 @@ const createWatcherState = (
   ...channelIds: string[]
 ): walletStates.Initialized => {
   const channelSubscriptions: walletStates.ChannelSubscriptions = {};
-  channelSubscriptions[processId] = channelIds;
+  for (const channelId of channelIds) {
+    if (!channelSubscriptions[channelId]) {
```
andrewgordstewart: `channelSubscriptions[channelId] = channelSubscriptions[channelId] || []`

packages/wallet/src/redux/protocols/virtual-funding/reducer.ts
```diff
@@ -234,12 +235,33 @@ function waitForGuarantorChannelReducer(
                 sharedData: indirectFundingResult.sharedData,
               };
             default:
+              const { targetChannelId, hubAddress, jointChannelId } = protocolState;
+              // We initialize our joint channel sub-protocol early in case we receive a commitment before we're done funding
```
andrewgordstewart: :+1:

packages/server/src/wallet/services/blockchain.ts
lalexgap: I'll add a more detailed comment. Out of sync might not be the best wording but I I believe this is the same issue you ran into here (a while ago): https://github.com/ethers-io/ethers.js/issues/363 where sending multiple transactions around the same time results in the error `the tx doesn't have the correct nonce. account has nonce of: 95 tx has nonce of: 94`

With virtual funding it's relatively easy to run into this since both player A and player B can deposit around the same time which results in the hub trying to do two deposits at roughly the same time.

My solution to avoid running into this error was to block so only one deposit occurs at a time.

lalexgap:
> It's hard to work out the reason for including `51e5dcac535107af8341b1b79ae1fe46c36d8fb2` in this PR. On the surface, it's hard to see why it's required in order to get virtual funding to work. It would be nice to have an explanation in the commit message.

The problem that this is solving is that the base funding reducer calls `routesToIndirectFunding` on every action it receives. Since Adjudicator Events had no `protocolLocator` this would always return true and the funding reducer would attempt to handle the event instead of properly routing it to virtual funding.

This could of fixed in the `fundingReducer` by some smarter routing based on its current state but this seemed like a more permanent solution that will avoid problems like this in the future.

PR #655, opened 2019-08-01
Virtual De-funding

### LINKED ISSUE #654: Virtual De-Funding Protocol
Add the protocol/tests/readme for the protocol that defunds a virtually funded channel.

lalexgap:
Adds a protocol to defund a virtually funded channel.

Fixes #654.

### DIFF
--- a/packages/wallet/src/redux/protocols/virtual-defunding/readme.md
+++ b/packages/wallet/src/redux/protocols/virtual-defunding/readme.md
@@ -1,0 +1,78 @@
+# Virtual De-Funding
+
+The virtual de-funding protocol takes a virtual channel that is funding an application channel and de-funds it resulting updated ledger channels with the hub.
+
+The virtual de-funding protocol starts with the following channels:
+
+- The application channel X which is now concluded.
+- The virtual channel J which funds X
+- Guarantor channels G0 (between player 0 and the hub) and G1(between player 1 and the hub)
+- Ledger channels L0(which funds G0) and L1(which funds G1)
+
+The virtual de-funding protocol does the following
+
+- Updates J to reflect the results of X
+- Updates L0 to reflect the results of J
+- Updates L1 to reflect the results of J
+
+After the protocol is run bother player 0 and player 1 have updated ledger channels with the hub that reflect the result of X.
+
+The virtual de-funding protocol leaves these ledger channels open so they can be re-used in the future or closed by another protocol.
+
+The Guarantor channels G0 and G1 as well as the joint channel J do not need to be updated and can simply be discarded after virtual de-funding.
+
+The virtual de-funding protocol involves three parties: player 0, player 1, and the hub H.
+Player 0 is identified by being the participant at index 0 in the `participants` array in
+the application channel X that is to be funded.
+
+In the following overview diagram, some communication related to embedded protocols are omitted.
+
+```mermaid
+sequenceDiagram
+  participant 0 as 0's wallet
+  participant H as Hub
+  participant 1 as 1's wallet
+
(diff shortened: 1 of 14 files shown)


packages/wallet/src/redux/protocols/virtual-defunding/readme.md
```diff
@@ -0,0 +1,78 @@
...
+
+- Updates J to reflect the results of X
+- Updates L0 to reflect the results of J
+- Updates L1 to reflect the results of J
+
+After the protocol is run bother player 0 and player 1 have updated ledger channels with the hub that reflect the result of X.
+
+The virtual de-funding protocol leaves these ledger channels open so they can be re-used in the future or closed by another protocol.
+
+The Guarantor channels G0 and G1 as well as the joint channel J do not need to be updated and can simply be discarded after virtual de-funding.
```
lalexgap: Currently `virtual-defunding` is not concluding the guarantor channels or joint channels. We should be concluding these so we have conclusion proofs on hand in case we want to get funds out.

packages/wallet/src/redux/protocols/virtual-defunding/reducer.ts
```diff
@@ -0,0 +1,188 @@
...
+import { getLatestCommitment, getTwoPlayerIndex } from '../reducer-helpers';
+import { addHex } from '../../../utils/hex-utils';
+import { VirtualDefundingAction } from './actions';
+import { routesToConsensusUpdate } from '../consensus-update/actions';
+import { HUB_ADDRESS } from '../../../constants';
+
+export function initialize({
+  processId,
+  targetChannelId,
+
```
andrewgordstewart:
```suggestion
```

packages/wallet/src/redux/protocols/virtual-defunding/reducer.ts
```diff
@@ -0,0 +1,188 @@
...
+  protocolLocator,
+  sharedData,
+}: {
+  processId: string;
+  targetChannelId: string;
+
```
andrewgordstewart:
```suggestion
```

packages/wallet/src/redux/protocols/virtual-defunding/reducer.ts
```diff
@@ -0,0 +1,188 @@
...
+        return {
+          protocolState: { ...protocolState, ledgerChannel },
+          sharedData,
+        };
+    }
+  }
+  return { protocolState, sharedData };
+}
+
+function getLedgerChannelId(jointChannelId: string, sharedData: SharedData): string {
```
andrewgordstewart: This name seems a bit ambiguous, because the joint channel whose ID you're passing in is already a ledger channel, but it's not the right ledger channel. Is there a name that makes it clear which ledger channel you're after?

packages/wallet/src/redux/state.ts
```diff
@@ -132,6 +132,7 @@ export interface FundingState {
 export interface ChannelFundingState {
```
andrewgordstewart: Should this now be split into a union type `DirectFundingState | IndirectFundingState | VirtualFundingState`? That would also clean up some of the checks required above on the existence of the `fundingChannel`

packages/wallet/src/redux/protocols/virtual-defunding/reducer.ts
lalexgap: This function gets refactored in another PR and has a more sensible name.

packages/wallet/src/redux/state.ts
lalexgap: Created #663 for this.

PR #658, opened 2019-08-07
Add Virtual De-Funding to de-funding protocol.

### LINKED ISSUE #657: Add Virtual-Defunding to Defunding protocol
The defunding protocol should handle a channel that is virtually funded and delegate to the virtual de-funding protocol.

lalexgap:
This gets virtual de-funding integrated into the parent defunding protocol.
It allows for the defunding/withdrawal from a game that has been virtually funded.

This is dependent on #655 #661, #660 and #659 and should only be merged after those.

### DIFF
--- a/packages/wallet/src/redux/protocols/virtual-funding/reducer.ts
+++ b/packages/wallet/src/redux/protocols/virtual-funding/reducer.ts
@@ -301,15 +301,7 @@ function waitForGuarantorFundingReducer(
   action: WalletAction,
 ) {
   const { processId, protocolLocator } = protocolState;
-  if (
-    !routesToIndirectFunding(action, protocolLocator) &&
-    !routesToConsensusUpdate(action, protocolLocator)
-  ) {
-    console.warn(
-      `Expected indirectFunding or consensusUpdate action, received ${action.type} instead`,
-    );
-    return { protocolState, sharedData };
-  }
+
   if (routesToConsensusUpdate(action, protocolLocator)) {
     let indirectApplicationFunding: consensusUpdate.ConsensusUpdateState;
     ({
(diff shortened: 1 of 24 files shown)


packages/wallet/src/redux/protocols/virtual-funding/reducer.ts
```diff
@@ -301,49 +301,80 @@ function waitForGuarantorFundingReducer(
...
+          protocolState: states.failure({
+            reason: 'Consensus Update failed or succeeded too early',
+          }),
+          sharedData,
+        };
+      default:
+        return { protocolState: { ...protocolState, indirectApplicationFunding }, sharedData };
+    }
+  }
+  const result = indirectFunding.indirectFundingReducer(
```
andrewgordstewart: Should this instead be
```
else if (!routesToIndirectFunding(action, protocolLocator)) { ... }
```
?

If you use some local enum like above, then the logic here is simpler.

packages/wallet/src/redux/protocols/virtual-funding/reducer.ts
```diff
@@ -301,49 +301,80 @@ function waitForGuarantorFundingReducer(
   action: WalletAction,
 ) {
   const { processId, protocolLocator } = protocolState;
-  if (routesToIndirectFunding(action, protocolLocator)) {
-    const result = indirectFunding.indirectFundingReducer(
-      protocolState.indirectGuarantorFunding,
+  if (
```
andrewgordstewart: Maybe this function could use a local enum like (ignore naming)
```
enum Direction { IndirectFunding, ConsensusUpdate, Other }
const direction: Direction =
  routesToIndirectFunding(action, protocolLocator) ? Direction.IndirectFunding :
  routesToConsensusUpdate(action, protocolLocator) ? Direction.ConsensusUpdate :
  Direction.Other

switch (direction) {
  ...
}
```

packages/wallet/src/redux/protocols/virtual-funding/reducer.ts
```diff
@@ -301,49 +301,80 @@ function waitForGuarantorFundingReducer(
...
+      default:
+        return { protocolState: { ...protocolState, indirectApplicationFunding }, sharedData };
+    }
+  }
+  const result = indirectFunding.indirectFundingReducer(
+    protocolState.indirectGuarantorFunding,
+    sharedData,
+    action,
+  );
+  if (indirectFunding.isTerminal(result.protocolState)) {
```
andrewgordstewart: Why not `switch (result.protocolState.type)` and put L369 in the default?

packages/wallet/src/redux/protocols/virtual-funding/reducer.ts
lalexgap: It looks like we don't get the type narrowing if we try to do it this way unfortunately.

packages/wallet/src/redux/protocols/virtual-funding/reducer.ts
lalexgap: That should be handled by the initial check ```  !routesToIndirectFunding(action, protocolLocator) &&
    !routesToConsensusUpdate(action, protocolLocator)``` if I understand your comment correctly.

packages/wallet/src/redux/protocols/direct-funding/reducer.ts
```diff
@@ -3,32 +3,40 @@ import { unreachable } from '../../../utils/reducer-utils';
 import { createDepositTransaction } from '../../../utils/transaction-generator';
 import * as actions from '../../actions';
 import { ProtocolReducer, ProtocolStateWithSharedData } from '../../protocols';
-import { SharedData } from '../../state';
+import { SharedData, registerChannelToMonitor } from '../../state';
```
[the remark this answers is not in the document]
lalexgap: I forgot that #660 was merged into this branch as well. The change should be from there.

packages/wallet/src/redux/protocols/virtual-funding/reducer.ts
andrewgordstewart: Good point. At the least, it would be good to use `if (expr1) { ... } else if (expr2) { ... }`, which is semantically equivalent?

packages/wallet/src/redux/protocols/direct-funding/actions.ts
```diff
@@ -1,36 +1,20 @@
 import * as actions from '../../actions';
-import { ActionConstructor } from '../../utils';
-import { TwoPartyPlayerIndex } from '../../types';
 import { isCommonAction, EmbeddedProtocol, routerFactory } from '../../../communication';
 
 // -------
 // Actions
 // -------
-export interface DirectFundingRequested {
```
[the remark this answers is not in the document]
lalexgap: I think that would make sense but should be unrelated to this change. Here I just switched `initialize` in `directFunding` to take named arguments instead of an action (so it follows the same pattern as other reducers)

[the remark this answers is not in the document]
andrewgordstewart: This comment was related to the fact that the `DirectFundingRequested` action was deleted.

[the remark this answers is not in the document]
lalexgap: `DirectFundingRequested` was only used to pass arguments into `direct-funding`'s `initialize` .

PR #659, opened 2019-08-07
Refactor Indirect-Defunding to use AdvanceChannel and ConsensusUpdate

lalexgap:
This PR does two things:
- Refactors the Indirect De-funding to use AdvanceChannel/ConsensusUpdate to handle sending/receiving commitments
- Add a `ClearedToSend` to allow indirectDefunding to be started early.

This was needed for work on #658

### DIFF
--- a/packages/wallet/src/redux/protocols/indirect-defunding/__tests__/scenarios.ts
+++ b/packages/wallet/src/redux/protocols/indirect-defunding/__tests__/scenarios.ts
@@ -35,14 +35,14 @@ export const clearedToSendHappyPath = {
       advanceChannelScenarios.postFund.preSuccess.sharedData,
     ),
     ...props,
-    clearedToSend: true,
+    clearedToProceed: true,
   },
   waitForLedgerUpdate: {
     state: waitForLedgerUpdate({
       ...props,
       ledgerUpdate,
       concluding,
-      ClearedToProceed: true,
+      clearedToProceed: true,
     }),
     action: ledgerUpdateScenarios.twoPlayerPreSuccessA.action,
     sharedData: ledgerUpdateScenarios.twoPlayerPreSuccessA.sharedData,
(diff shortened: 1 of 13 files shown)


packages/wallet/src/redux/protocols/indirect-defunding/__tests__/scenarios.ts
```diff
@@ -23,179 +12,45 @@ const twoThree = [
...
-const ledger6 = ledgerCommitment({ turnNum: 6, balances: fiveToApp, proposedBalances: twoThree });
-export const ledger7 = ledgerCommitment({ turnNum: 7, balances: twoThree });
-const ledger8 = ledgerCommitment({ turnNum: 8, balances: twoThree, isFinal: true });
-const ledger9 = ledgerCommitment({ turnNum: 9, balances: twoThree, isFinal: true });
-
-// -----------
-// States
-// -----------
+const ledgerUpdate = ledgerUpdateScenarios.twoPlayerPreSuccessA.state;
+// TODO: We should probably add a conclude scenario to advance channel
```
andrewgordstewart:
```suggestion
```
Make this into a zube card?

packages/wallet/src/redux/protocols/indirect-defunding/actions.ts
```diff
@@ -1,18 +1,45 @@
...
+  routerFactory,
+  EmbeddedProtocol,
+  BaseProcessAction,
+  ProtocolLocator,
+} from '../../../communication';
+import { ActionConstructor } from '../../utils';
 // -------
 // Actions
 // -------
+export interface ClearedToSend extends BaseProcessAction {
```
andrewgordstewart: Is this more like `ClearedToProceed`? Is this idea that
- the protocols should be started up in order to receive messages;
- but the user might not have ok'd the defunding yet
- and once the user clears it, all sub-protocols are then cleared to send?

If that's the case, I think `ClearedToProceed` or `ClearedToDefund` might be a tad clearer, since the de-funding protocol itself doesn't actually send anything.

lalexgap: renamed to `clearedToProceed`

PR #660, opened 2019-08-08
Fix Funding Race Condition

lalexgap:
Fixed some race conditions that caused the funding protocol to stall.
- Handled the Funding Received Event coming in early in the Direct Funding protocol
- Handled the app funding ledger commitment coming in early in the Virtual Funding protocol

Also did some minor refactoring in Direct Funding so it takes arguments in initialize instead of an action (to follow the pattern of other `initialize` methods)


lalexgap:
Closing this as it got merged as part of #658

PR #661, opened 2019-08-08
Add Conclude support to AdvanceChannel

lalexgap:


packages/wallet/src/redux/protocols/advance-channel/__tests__/scenarios.ts
```diff
@@ -313,6 +352,29 @@ export const existingChannelAsA = {
   },
 };
 
+// TODO: Add tests for concluding as B and hub
```
andrewgordstewart: Can you make this a zube card rather than a TODO?

andrewgordstewart:
```suggestion
```
Can you make this a zube card rather than a TODO?

packages/wallet/src/redux/protocols/advance-channel/reducer.ts
```diff
@@ -315,3 +311,39 @@ function channelAdvanced(channel: ChannelState, commitmentType: CommitmentType):
...
+  } else {
+    const next = nextSetupCommitment(lastCommitment);
+    if (next === 'NotASetupCommitment') {
+      throw new Error('lastCommitment was not a setup commitment');
+    }
+    return next;
+  }
+}
+
+// This only works for 2 or 3 players
```
andrewgordstewart: Maybe throw an error if the parameters of the channel don't match this requirement, rather than having this comment?

packages/wallet/src/redux/protocols/advance-channel/reducer.ts
```diff
@@ -315,3 +311,39 @@ function channelAdvanced(channel: ChannelState, commitmentType: CommitmentType):
     lastCommitment.commitmentType > commitmentType
   );
 }
+
+function nextCommitment(channel: ChannelState, commitmentType: CommitmentType): Commitment {
+  const lastCommitment = getLastCommitment(channel);
```
andrewgordstewart: Can't you just look at the round of commitments that is stored, count how many have the correct commitment type, and add one?

```
const commitments = getCommitments(channel); // fake method
const commitmentCount = commitments.find(c => c.commitmentType === commitmentType).length + 1;
```

Issue #559, opened 2019-06-27
Add lint rule to disallow single letter variables


andrewgordstewart:
```suggestion
export const waitForChannelPreparation: StateConstructor<WaitForChannelPreparation> = params => {
```

_Originally posted by @kerzhner in https://github.com/magmo/apps/pull/554_

PR #667, opened 2019-08-16
Rework Concluding/Defunding

lalexgap:
This reworks defunding and concluding to prepare us for long lived channels.

Some of the big changes:
 - Defunding no longer attempts to withdraw, it just updates ledger channels to allocate funds back to the players. This way the ledger channel can be re-used.
- A new protocol `close-ledger-channel` has been added which is responsible for concluding a ledger channel, and withdrawing the funds from the channel.
- Concluding has been refactored to use these protocols and been slimmed down to one reducer instead of two.


packages/wallet/src/redux/protocols/concluding/reducer.ts
```diff
@@ -1,44 +1,238 @@
...
+
+export function initialize({
+  channelId,
+  processId,
+  opponentInstigatedConclude,
+  sharedData,
+}: {
+  channelId: string;
+  processId: string;
+  opponentInstigatedConclude: boolean;
```
andrewgordstewart: Does it make more sense for this to be `weInstigatedConclude`?

packages/wallet/src/redux/protocols/concluding/states.ts
```diff
@@ -1,24 +1,51 @@
...
+  reason: string;
 }
 
 export interface Success {
   type: 'Concluding.Success';
 }
 
+export interface WaitForConclude {
+  type: 'Concluding.WaitForConclude';
+  channelId: string;
```
andrewgordstewart: Can you use a base interface with `channelId`, `processId`, and `ledgerId`?

packages/wallet/src/redux/protocols/defunding/actions.ts
```diff
@@ -16,12 +17,14 @@ import { VirtualDefundingAction, isVirtualDefundingAction } from '../virtual-def
 // -------
 
 // TODO: Replace once ledger defunding actions are defined
```
andrewgordstewart: Have the ledger defunding actions now been defined?

packages/wallet/src/redux/protocols/ledger-defunding/states.ts
```diff
@@ -0,0 +1,65 @@
...
+  ledgerId: string;
+  channelId: string;
+  clearedToProceed: boolean;
+  ledgerUpdate: ConsensusUpdateState;
+  protocolLocator: ProtocolLocator;
+}
+
+export interface Failure {
+  type: 'LedgerDefunding.Failure';
+  reason: string;
```
andrewgordstewart: Should we enumerate the possible failure reasons?

packages/wallet/src/redux/protocols/reducer-helpers.ts
```diff
@@ -11,7 +11,9 @@ import * as magmoWalletClient from 'magmo-wallet-client';
 import { getLastCommitment, nextParticipant, Commitments } from '../channel-store';
 import { Commitment } from '../../domain';
 import { sendCommitmentsReceived, ProtocolLocator } from '../../communication';
+import * as comms from '../../communication';
 import { ourTurn as ourTurnOnChannel } from '../channel-store';
+import _ from 'lodash';
```
andrewgordstewart: Is this introducing `lodash` as a dependency? I only recall seeing it in test files.

packages/wallet/src/redux/protocols/reducer-helpers.ts
```diff
@@ -174,6 +186,13 @@ export const channelIsClosed = (channelId: string, sharedData: SharedData): bool
   );
 };
 
+export const channelFundsAnotherChannel = (channelId: string, sharedData: SharedData): boolean => {
+  const latestCommitment = getLatestCommitment(channelId, sharedData);
+  return (
+    _.intersection(selectors.getChannelIds(sharedData), latestCommitment.destination).length > 0
```
andrewgordstewart:
```suggestion
    !!_.intersection(selectors.getChannelIds(sharedData), latestCommitment.destination)
```

packages/wallet/src/redux/protocols/close-ledger-channel/reducer.ts
```diff
@@ -0,0 +1,154 @@
...
+  if (!isWithdrawalAction(action)) {
+    return { protocolState, sharedData };
+  }
+  let withdrawal: WithdrawalState;
+  ({ protocolState: withdrawal, sharedData } = withdrawalReducer(
+    protocolState.withdrawal,
+    sharedData,
+    action,
+  ));
+  if (withdrawal.type === 'Withdrawing.Success') {
```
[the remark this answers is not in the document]
lalexgap: I think this was moved from existing code, so it has some older patterns

packages/wallet/src/redux/protocols/reducer-helpers.ts
lalexgap: We're making use of `lodash` in a couple of (non-test) places. Is there an issue having `lodash` as a dependency?

packages/wallet/src/redux/protocols/defunding/actions.ts
lalexgap: yup, removed

packages/wallet/src/redux/protocols/close-ledger-channel/reducer.ts
```diff
@@ -0,0 +1,154 @@
...
+      }),
+      sharedData,
+    };
+  }
+};
+
+const createWaitForWithdrawal = (sharedData: SharedData, processId: string, channelId: string) => {
+  const withdrawalAmount = getWithdrawalAmount(sharedData, channelId);
+
+  const { protocolState: withdrawal, sharedData: newSharedData } = withdrawalInitialize(
```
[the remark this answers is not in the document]
lalexgap: I think this was moved from existing code, so it has some older patterns

packages/wallet/src/redux/protocols/close-ledger-channel/reducer.ts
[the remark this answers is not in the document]
andrewgordstewart: Should we take the opportunity to use the new patterns then?

packages/wallet/src/redux/protocols/close-ledger-channel/states.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export interface Success {
+  type: 'CloseLedgerChannel.Success';
+}
+
+// -------
+// Constructors
+// -------
+
+export const waitForWithdrawal: StateConstructor<WaitForWithdrawal> = p => {
```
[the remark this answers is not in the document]
lalexgap: Are we committed to disallowing single letter variables? (https://github.com/magmo/apps/issues/559)

I feel like it can be pretty readable (and is something we do all over the place)

packages/wallet/src/redux/protocols/close-ledger-channel/reducer.ts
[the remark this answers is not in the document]
lalexgap: Yup, updated.

packages/wallet/src/redux/protocols/close-ledger-channel/reducer.ts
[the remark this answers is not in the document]
lalexgap: Updated

packages/wallet/src/redux/protocols/close-ledger-channel/states.ts
```diff
@@ -0,0 +1,71 @@
...
+export interface WaitForConclude {
+  type: 'CloseLedgerChannel.WaitForConclude';
+  processId: string;
+  concluding: AdvanceChannelState;
+  channelId;
+}
+
+export interface Failure {
+  type: 'CloseLedgerChannel.Failure';
+  reason: string;
```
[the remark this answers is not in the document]
lalexgap: I've moved away from doing this just for ease of development and the fact that we don't use the reason yet.

It might be good for future proofing so I'll go through and update them.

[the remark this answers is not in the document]
andrewgordstewart: I think it has these benefits:
- it serves as good documentation
- it enables auto-complete
- it will be useful when specifying how protocols should handle failure

packages/wallet/src/redux/protocols/concluding/reducer.ts
lalexgap: I think either make sense to me, is there a reason you prefer `weInstigatedConclude` ?

packages/wallet/src/redux/protocols/reducer-helpers.ts
lalexgap: Just a heads up that this doesn't work since `!![]` evaluates to `true`

andrewgordstewart: Right, that was a bit silly. !![].length does work, though I don't know if there's much benefit of doing that.

PR #668, opened 2019-08-17
Remove process when its done

### LINKED ISSUE #653: Processes are not removed when completed
This is relatively minor as it doesn't cause any problems.

Currently after a protocols is completed it is not removed from the state, and neither are its adjudicator event subscriptions. It'd be nice if these were cleaned up on completion. It currently causes warnings in the console as adjudicator events are still dispatched to the completed process.

lalexgap:
Fixes #653.

Removes processes when they're done.

### DIFF
--- a/packages/wallet/src/redux/__tests__/initialized.test.ts
+++ b/packages/wallet/src/redux/__tests__/initialized.test.ts
@@ -91,3 +91,34 @@ describe('when a updateSharedData action arrives', () => {
     expect(reducer).toHaveBeenCalledWith(state.adjudicatorState, action);
   });
 });
+
+describe('when a process state is terminal', () => {
+  const processId = '0xprocessId';
+  const protocolState = {};
+  const processState: states.ProcessState = {
+    processId,
+    protocol: ProcessProtocol.Funding,
+    channelsToMonitor: [],
+    protocolState,
+  };
+  const state = { ...initializedState, processStore: { [processId]: processState } };
+  const action = strategyApproved({
+    processId: '0xprocessId',
+    strategy: 'IndirectFundingStrategy',
+  });
+  const reducer = jest.fn(() => ({
+    protocolState: { type: 'Funding.Success' },
+    sharedData: 'sharedData ',
+  }));
+  Object.defineProperty(fundProtocol, 'fundingReducer', {
+    value: reducer,
+  });
+
+  const result = walletReducer(state, action);
+  it('removes the current process id', () => {
+    expect(result.currentProcessId).toBeUndefined();
+  });
+  it('removes the process state', () => {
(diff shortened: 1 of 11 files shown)


packages/wallet/src/redux/state.ts
```diff
@@ -322,3 +335,13 @@ export function getCommitments(store: SharedData, channelId: string): Commitment
...
 
 export { NewLedgerChannel };
+
+export function isTerminalProcessProtocolState(
+  protocolState: ProtocolState,
+): protocolState is TerminalApplicationState | TerminalFundingState | TerminalDefundingState {
+  return (
+    (isApplicationState(protocolState) && isTerminalApplicationState(protocolState)) ||
+    (isFundingState(protocolState) && isTerminalFundingState(protocolState)) ||
+    (isDefundingState(protocolState) && isTerminalDefundingState(protocolState))
```
geoknee: What about the `TerminalConcludingState` case?

packages/wallet/src/redux/state.ts
```diff
@@ -322,3 +335,13 @@ export function getCommitments(store: SharedData, channelId: string): Commitment
 }
 
 export { NewLedgerChannel };
+
+export function isTerminalProcessProtocolState(
```
geoknee: Is `isTerminalProtocolState` a better name?

packages/wallet/src/redux/state.ts
lalexgap: Oops, fixed!

packages/wallet/src/redux/state.ts
lalexgap: Renamed

packages/wallet/src/redux/__tests__/initialized.test.ts
```diff
@@ -91,3 +91,34 @@ describe('when a updateSharedData action arrives', () => {
...
+    protocolState: { type: 'Funding.Success' },
+    sharedData: 'sharedData ',
+  }));
+  Object.defineProperty(fundProtocol, 'fundingReducer', {
+    value: reducer,
+  });
+
+  const result = walletReducer(state, action);
+  it('removes the current process id', () => {
+    expect(result.currentProcessId).toBeUndefined();
```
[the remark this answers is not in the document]
lalexgap: I think we'll need to think further about handling concurrent running processes so when we make those changes we can probably change this.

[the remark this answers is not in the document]
lalexgap: I've created #669 to address handling concurrent processes.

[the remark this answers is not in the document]
geoknee: 👍

PR #670, opened 2019-08-20
Channel Management 

lalexgap:
Introduces a rudimentary UI that shows open channels and should allow you to close an open channel with the hub.


packages/wallet/src/redux/protocols/channel-management/reducer.ts
```diff
@@ -0,0 +1,84 @@
...
+    ourAddress: getOurAddress(channelId, sharedData),
+    opponentAddress: getOpponentAddress(channelId, sharedData),
+    ourAmount: getOurAllocation(channelId, sharedData),
+    opponentAmount: getOpponentAllocation(channelId, sharedData),
+    channelAmount: getAllocationTotal(channelId, sharedData),
+    fundingChannel: fundingChannel === channelId ? '' : fundingChannel,
+    fundedBy,
+  };
+};
+export const initialize = ({
```
lalexgap: I'm not sure if using a protocol is the best way to handle something that is mostly UI driven. It seems against the "engine" pattern we've been discussing.

lalexgap:
Closing this to to focus on supporting the new `nitro-protocol`.

PR #672, opened 2019-08-21
Fix Existing Ledger Channel Funding

### LINKED ISSUE #671: Re-Using Existing Ledger Channel Does not Preserve Funds
Whenever you fund a game using an existing ledger channel, it will not preserve funds properly.

For example if you have L=[A:10,B:10] and want to fund X=[A:5,B:5] you end up with L=[X:10], which loses 10 eth. Instead we should end up with L=[X:5,A:5,B:5].

lalexgap:
Fixes #671


packages/wallet/src/redux/protocols/existing-ledger-funding/reducer.ts
```diff
@@ -265,12 +266,42 @@ function ledgerChannelNeedsTopUp(
...
+  const appTotal = startingAllocation.reduce(addHex);
+
+  // If the ledger allocation is greater than the startingAllocation requested
+  // we subtract the startingAllocation from the ledger allocation
+  const updatedLedgerAllocation = ledgerAllocation.map(a => {
+    const index = ledgerAllocation.indexOf(a);
+    const updatedEntry = subHex(a, startingAllocation[index]);
+
+    if (bigNumberify(updatedEntry).lt(0)) {
+      throw new Error(
```
andrewgordstewart: Should the callers of `craftAppFunding` catch this error, and move to a failure state?

packages/wallet/src/redux/protocols/reducer-helpers.ts
```diff
@@ -321,3 +323,18 @@ export function getFundingChannelId(channelId: string, sharedData: SharedData):
...
   }
 }
+
+export function removeZeroFundsFromBalance(
+  incomingAllocation: string[],
+  incomingDestination: string[],
+): { allocation: string[]; destination: string[] } {
+  const allocation: string[] = [];
+  const destination: string[] = [];
+  for (let i = 0; i < incomingAllocation.length; i++) {
```
andrewgordstewart: This could be something like
```
incomingAllocation.map((a, i) => {
  if (bigNumberify(a).gt(0)) {
      allocation.push(incomingAllocation[i]);
      destination.push(incomingDestination[i]);
  }
}
```

packages/wallet/src/redux/protocols/existing-ledger-funding/reducer.ts
lalexgap: This check was incorrect (as a ledger channel can be underfunded and will be topped up). I've updated the logic here.

PR #673, opened 2019-08-21
Handle invalid app transitions

lalexgap:
If an app transition was invalid we would throw an error trying to decode the appAttributes of non-consensus rules.

### DIFF
--- a/packages/wallet/src/redux/sagas/message-listener.ts
+++ b/packages/wallet/src/redux/sagas/message-listener.ts
@@ -12,6 +12,7 @@ import * as contractUtils from '../../utils/contract-utils';
 import { concluded, challengeRequested } from '../protocols/application/actions';
 
 import { appAttributesFromBytes } from 'fmg-nitro-adjudicator/lib/consensus-app';
+import { CONSENSUS_LIBRARY_ADDRESS } from '../../constants';
 export function* messageListener() {
   const postMessageEventChannel = eventChannel(emitter => {
     window.addEventListener('message', (event: MessageEvent) => {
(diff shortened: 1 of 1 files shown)


packages/wallet/src/redux/sagas/message-listener.ts
```diff
@@ -149,8 +150,13 @@ function* validateAgainstLatestCommitment(incomingCommitment: Commitment) {
 function* validateTransition(fromCommitment: Commitment, toCommitment: Commitment) {
   const validTransition = yield contractUtils.validateTransition(fromCommitment, toCommitment);
   if (!validTransition) {
-    const fromAppAttributes = appAttributesFromBytes(fromCommitment.appAttributes);
-    const toAppAttributes = appAttributesFromBytes(toCommitment.appAttributes);
+    const isConsensusChannel = toCommitment.channel.channelType === CONSENSUS_LIBRARY_ADDRESS;
+    const fromAppAttributes = isConsensusChannel
+      ? appAttributesFromBytes(fromCommitment.appAttributes)
```
andrewgordstewart: It might be worth having a decoder lookup defined somewhere, so that this could be
`fromAppAttributes = decoder[fromCommitment.channel.channelType](fromCommitment.appAttributes)`

lalexgap: Would there be a case where we expect the wallet to decode anything other than consensus app attributes? Unless the wallet is is provided a decode function from an app, it should only every try this on the consensus rules.

If that's the case I think it's nicer to have it inlined so you understand the exact behaviour from looking at this function.

andrewgordstewart: True, good point!

PR #674, opened 2019-08-23
Switch currentProcessId to a queue

### LINKED ISSUE #669: Handle Process scheduling
To support having multiple running channels we need some kind of way of scheduling processes, so it is possible to switch from process to another.

Instead of just using `currentProcessId` we probably want some kind of priority queue.

lalexgap:
Switches the `currentProcessId` to queue of `processId`s sorted by`priority`.

Fixes #669.


lalexgap:
Closing this to to focus on supporting the new `nitro-protocol`.

PR #685, opened 2019-09-24
Small code cleanup to WalletInitializedContainer

snario:


lalexgap (approved):

packages/wallet/src/containers/initialized.tsx
```diff
@@ -12,12 +12,12 @@ interface Props {
 class WalletInitializedContainer extends PureComponent<Props> {
   render() {
     const { state } = this.props;
-    if (!state.currentProcessId) {
-      return <LandingPage />;
-    } else {
+
```
lalexgap, line 15: I think prettier needs to be run on this file. You can do `yarn prettier:write` or just use the `prettier` vs code extension.

